PR #40, opened 2015-03-02
Fix Maven build warning from xml-apis

rwhogg:
xml-apis version 2.0.2 was renamed (for reasons unknown) to 1.0.b2, which causes a Maven warning when doing ```mvn install```.. This change fixes that.


sideshowbarker:
Thanks!

PR #43, opened 2015-03-03
Removes all unused imports from the Java files in the src/ directory

rwhogg:
This removes all the unused Java imports from the src/ directory. They were found with PMD.

### DIFF
--- a/src/nu/validator/gnu/xml/aelfred2/XmlParser.java
+++ b/src/nu/validator/gnu/xml/aelfred2/XmlParser.java
@@ -77,9 +77,8 @@ import nu.validator.htmlparser.io.Encoding;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
-import com.ibm.icu.text.UnicodeSet;
-
 // Organized imports -- 2005-08-20 hsivonen
+// Removed unused imports -- 2015-03-03 rwhogg
 
 /**
  * Parse XML documents and return parse events through call-backs. Use the
--- a/src/nu/validator/localentities/LocalCacheEntityResolver.java
+++ b/src/nu/validator/localentities/LocalCacheEntityResolver.java
@@ -1,8 +1,6 @@
 package nu.validator.localentities;
 
 import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
--- a/src/nu/validator/servlet/Html5ConformanceCheckerTransaction.java
+++ b/src/nu/validator/servlet/Html5ConformanceCheckerTransaction.java
@@ -36,7 +36,6 @@ import org.xml.sax.SAXNotRecognizedException;
 import org.xml.sax.SAXNotSupportedException;
 import org.xml.sax.SAXParseException;
 
-import com.thaiopensource.relaxng.impl.CombineValidator;
 import com.thaiopensource.validate.IncorrectSchemaException;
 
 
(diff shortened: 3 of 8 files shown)


rwhogg:
Two side notes:
1. Do you want me to fill in the CLA? I already affirmed last year, before you had the CLAHub process in place, but I can do it again if you need me to.
2. Would you be interested in a more-substantial PMD refactor of the codebase? Alternatively, are there any classes in particular that need cleanup? (The GNU classes looked like definite candidates to me)

sideshowbarker:
> 1. Do you want me to fill in the CLA? I already affirmed last year, before you had the CLAHub process in place, but I can do it again if you need me to.

Thanks for reminding me about that. If you could sign https://www.clahub.com/agreements/validator/validator that'd be great but actually I'd like to move to a "Signed-of-by" process. So if you can read https://raw.githubusercontent.com/validator/validator/master/DCO and confirm that you agree, please sign any future commit messages by doing a `commit -s` to add the "Signed-of-by" line (or just add it yourself manually).

In fact if you could amend your message for this commit to add that line I'd appreciate it. But if not I can just merge it this time as-is.

sideshowbarker:
> 2. Would you be interested in a more-substantial PMD refactor of the codebase? Alternatively, are there any classes in particular that need cleanup? (The GNU classes looked like definite candidates to me)

I would be happy to look at the results and review them as long as it's not something that requires a lot of manual work from you. Because I'm not sure yet whether we'd definitely want to land the results or not (I'd need to see what they look like first) so I wouldn't want you to put a bunch of time into it and then end up with it being wasted effort.

I've looked at the results from some of the simpler refactoring fixup stuff that Eclipse provides off the shelf but I've never felt it was important enough to me to actually land it. And I've never run PMD on anything myself but I can imagine that while it produces more substantially helpful results, it might also produce a pretty large volume of them and require significant time to review.

But that said, I know other potential contributors care a lot about some of this stuff, and also it's probably better to get this kind of cleanup done now—before more contributors show up and get involved—than it would be do it later, when it would be more disruptive to a larger number of people.

rwhogg:
I signed-off the commit. I tried to use CLAHub too, but it always says "Something went wrong" after I try to sign.

I don't have the time to investigate why right now, but I can later if you need.

sideshowbarker:
> I signed-off the commit. I tried to use CLAHub too, but it always says "Something went wrong" after I try to sign.

Thanks

> I don't have the time to investigate why right now, but I can later if you need.

No worries. I've also found CLAHub to be unreliable so I reckon it's probably better at this poin to just shut it off so any later contributors don't need to bother with it, and instead just switch completely to using a signed-off-by system.

sideshowbarker:
> 2. Would you be interested in a more-substantial PMD refactor of the codebase? Alternatively, are there any classes in particular that need cleanup? (The GNU classes looked like definite candidates to me)

The first time I read your message I missed the fact you'd mentioned the GNU classes. I assume you mean the https://github.com/validator/validator/tree/master/src/nu/validator/gnu/xml/aelfred2 tree.

If so, to be clear, I don't think we'd want to land any changes to that code other than bug fixes. The reason is that it's just a fork of old third-party code that was brought in with a few relatively small changes to ease integration of it for use in the validator codebase. It's not code that any contributor is likely to need or want to hack further on going forward, so the value of refactoring it is relatively minimal.

That said, I still think there could be some benefit to looking at a PMD refactor of code in the rest of the src tree that we're actively maintaining.

rwhogg:
Good point. I'll see what I can do with some of the other classes. Unfortunately, some of it might take a bit of manual work, but I'll maybe do it in smaller chunks so we can decide over time what changes are worth making.

sideshowbarker:
Sounds good

PR #44, opened 2015-03-04
Various minor refactors:

rwhogg:
- some more unused imports removed
- some unused fields removed
- a frequently-repeated long string made into a constant
- some performance improvements

### DIFF
--- a/src/nu/validator/checker/UnsupportedFeatureChecker.java
+++ b/src/nu/validator/checker/UnsupportedFeatureChecker.java
@@ -11,8 +11,6 @@ public class UnsupportedFeatureChecker extends Checker {
      */
     @Override public void startElement(String uri, String localName,
             String qName, Attributes atts) throws SAXException {
-        boolean followW3Cspec = "1".equals(System.getProperty("nu.validator.servlet.follow-w3c-spec")) ? true
-            : false;
         if ("http://www.w3.org/1999/xhtml" != uri) {
             return;
         }
--- a/src/nu/validator/checker/XmlPiChecker.java
+++ b/src/nu/validator/checker/XmlPiChecker.java
@@ -1377,7 +1377,7 @@ public class XmlPiChecker extends Checker implements LexicalHandler {
     private void addAttributeWithValue() throws SAXException {
         if (attributeName != null) {
             String value = longStrBufToString();
-            if (value.indexOf("<") != -1) {
+            if (value.indexOf('<') != -1) {
                 errAttributeValueContainsLt();
                 return;
             }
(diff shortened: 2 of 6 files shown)


src/nu/validator/checker/UnsupportedFeatureChecker.java
```diff
@@ -11,8 +11,6 @@
      */
     @Override public void startElement(String uri, String localName,
             String qName, Attributes atts) throws SAXException {
-        boolean followW3Cspec = "1".equals(System.getProperty("nu.validator.servlet.follow-w3c-spec")) ? true
-            : false;
         if ("http://www.w3.org/1999/xhtml" != uri) {
```
rwhogg: This same URL appears in at least 32 different files. Maybe we should consider creating a constants class to hold it? Assuming there are enough strings with similar frequency to make it worthwhile.

src/nu/validator/checker/XmlPiChecker.java
```diff
@@ -1377,7 +1377,7 @@ private void attributeNameComplete() throws SAXException {
     private void addAttributeWithValue() throws SAXException {
         if (attributeName != null) {
             String value = longStrBufToString();
-            if (value.indexOf("<") != -1) {
+            if (value.indexOf('<') != -1) {
```
rwhogg: This is just a performance improvement - probably a very small one, but it's no less readable.

src/nu/validator/datatype/Charset.java
```diff
@@ -22,11 +22,6 @@
 
 package nu.validator.datatype;
 
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
-
```
rwhogg: I guess I was wrong about getting rid of ALL the unused imports with the last commit. Whoops! No harm done at least.

src/nu/validator/source/Location.java
```diff
@@ -30,8 +30,6 @@
  * @author hsivonen
  */
 public final class Location implements Comparable<Location>, Cloneable {
-
-    private static int compares = 0;
```
rwhogg: This variable (previously was) added to down below, but is never actually read.

src/nu/validator/checker/UnsupportedFeatureChecker.java
sideshowbarker: > This same URL appears in at least 32 different files. Maybe we should consider creating a constants class to hold it? Assuming there are enough strings with similar frequency to make it worthwhile.

A grep shows there's 163 instances of it in code in the tree so yeah it might make some sense.

src/nu/validator/checker/XmlPiChecker.java
sideshowbarker: > This is just a performance improvement - probably a very small one, but it's no less readable.

Sure yeah it should rightly have been written this way to begin with, so it's nice to get it changed now.

src/nu/validator/datatype/Charset.java
sideshowbarker: > I guess I was wrong about getting rid of ALL the unused imports with the last commit. Whoops! No harm done at least.

Yeah, no worries

src/nu/validator/source/Location.java
sideshowbarker: > This variable (previously was) added to down below, but is never actually read.

Yup, no need to keep it around

PR #45, opened 2015-03-04
Removes warning about having to build twice the first time

rwhogg:
According to #10 building twice is no longer actually necessary.


sideshowbarker:
yeah actually I'd forgotten I had fixed this :-)

If you would please also update the `.travis.yml` file so there's only one `python ./build/build.py build` invocation on line 5 I'd appreciate it.

rwhogg:
np, done

PR #46, opened 2015-03-05
Various small refactors of the XHTML handling code

rwhogg:


sideshowbarker:
btw this is a super-minor nit, but as far as the change description here, the affected code’s not so really “XHTML handling code“ but instead more like “error-handling code for emitting HTML-formatted error messages“. All the `messages/Xhtml*` classes could just as well be named `messages/Html*` instead—there's actually nothing XHTML-specific about them at all and you’ll notice no corresponding `Html*` classes actually already exist there.

What this code really is, it's the core main code that handles the messages that most users of the validator server actually see—in that it's the code that formats the messages to display to users in the Web UI of the validator service.

PR #48, opened 2015-03-06
Remove unnecessary finals and returns

rwhogg:

We're starting to run out of minimal-effort, obviously-beneficial refactors. So for this PR, you might or might not agree that the changes are worth doing. Fortunately they didn't take long to do.

PMD has a way to design custom rulesets, and I'm going to try to make one later tonight. Whether or not you accept everything in this PR will affect what goes into it. The plan is to commit that ruleset and add a section to the Ant script to run it, so we can track whether the code base is actually improving over time.

### DIFF
--- a/src/nu/validator/checker/table/HorizontalCellComparator.java
+++ b/src/nu/validator/checker/table/HorizontalCellComparator.java
@@ -38,7 +38,7 @@ final class HorizontalCellComparator implements Comparator<Cell> {
         super();
     }
 
-    public final int compare(Cell cell0, Cell cell1) {
+    public int compare(Cell cell0, Cell cell1) {
         if (cell0.getLeft() < cell1.getLeft()) {
             return -1;
         } else if (cell0.getLeft() > cell1.getLeft()) {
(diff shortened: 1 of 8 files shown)


sideshowbarker:
It would be good to have @hsivonen weigh in with an opinion both on this particular changeset, and on the general question around what set of rules we'd want in a PMD custom ruleset, if we were to agree to have one.

IMHO for the purpose of helping enforce code consistency among contributed code, there's some value in having a published PMD custom ruleset that any contributor could run on their code before submitting a PR—and that any reviewers could run on it before merging the PR.

That said, I don't feel super strongly about it, so if @hsivonen reckons it wouldn't be worth it, then I'd defer to him on it.

sideshowbarker:
BTW as far as the substance of this particular changeset IMHO there is value in removing unnecessary finals especially. Among other reasons, at the simplest level, it can help somebody learning the code understand when `final` is necessary and when it's not—and help to avoid confusing them by having it in places where it's not actually necessary.

PR #54, opened 2015-03-09
More tiny refactors

rwhogg:
Not much here that's too exciting, but I guess that's just a sign of a good clean codebase.

### DIFF
--- a/src/nu/validator/checker/table/VerticalCellComparator.java
+++ b/src/nu/validator/checker/table/VerticalCellComparator.java
@@ -41,7 +41,7 @@ final class VerticalCellComparator implements Comparator<Cell> {
         super();
     }
 
-    public final int compare(Cell cell0, Cell cell1) {
+    public int compare(Cell cell0, Cell cell1) {
         if (cell0.getBottom() < cell1.getBottom()) {
             return -1;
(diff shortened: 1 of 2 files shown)


sideshowbarker:
Keep ’em coming. I tried running PMD myself but even the “basic” ruleset returns a lot of noise. It’d be nice to try to put together a common ruleset we could agree on. But I’d definitely want @hsivonen to weigh in.

sideshowbarker:
btw, e.g., what PMD rule looks for unnecessary `final`? or for use of double quotes around single characters (where using single quotes would be best practice and more performant).

rwhogg:
No kidding about the noise. The one that annoys me the most is the "unnecessary parentheses" rule.

http://pmd.sourceforge.net/pmd-5.2.3/pmd-java/rules/java/strings.html#UseIndexOfChar is what I've been using to find double quotes around single characters - it's only for indexOf() though, not a general check.

http://pmd.sourceforge.net/pmd-5.2.3/pmd-java/rules/java/unnecessary.html#UnnecessaryFinalModifier checks for unnecessary finals inside of an already final class.

sideshowbarker:
ah ok, thanks for the pointers

PR #64, opened 2015-03-11
Some small refactors in the checker code

rwhogg:


sideshowbarker:
Nit: Please always use four spaces instead of tabs.

Also if you’re not using Eclipse, please take a look at the formatting rules here:
https://github.com/rwhogg/validator/blob/master/.settings/org.eclipse.jdt.core.prefs

I don't know if that pref file is exportable for use in other IDEs, but short of that if you can configure your IDE to match those as well as you can, it would be appreciated.

rwhogg:
I apologize. I've told Emacs a thousand times to use spaces instead of tabs, and every time I try something even slightly out of the ordinary with it, it starts using tabs on me instead.

sideshowbarker:
No need to apologize :-) It didn’t create any real problem

sideshowbarker:
FYI I added a portable file to the repo that contains eclipse formatter settings.
https://github.com/validator/validator/commit/3689b23e6e9ab69aab3c8a8ef86a3a4b36d358af

It’s supposedly possible to import that file into a NetBeans project—
http://plugins.netbeans.org/plugin/50877/eclipse-code-formatter-for-java
https://github.com/markiewb/eclipsecodeformatter_for_netbeans

PR #77, opened 2015-03-18
Makes classes final.

rwhogg:
Some API changes here, but not any that make any substantive difference. I just changed classes that had private constructors to be final.


src/nu/validator/datatype/DatetimeLocal.java
```diff
@@ -59,7 +59,7 @@ private DatetimeLocal() {
      * @return the regexp for this datatype
      * @see nu.validator.datatype.AbstractDatetime#getPattern()
      */
-    protected final Pattern getPattern() {
+    protected Pattern getPattern() {
```
rwhogg: One exception to the commit title...

PR #81, opened 2015-03-30
For file uploads, use supplied filename as sysID.

sideshowbarker:
Form-based file uploads sent as `multipart/form-data` typically include a
field that gives the original filename of the uploaded file. This change
causes the filename to be included in every error or warning message the
checker emits when checking the uploaded file.

Note that there's an important limitation to understand for this case: The
filename information made available to the backend for file uploads is
simply just a filename -- not an absolute pathname and not even a relative
pathname. That means, e.g., if you're recursing through a directory tree,
the messages emitted for a file named `index.html` at the top level and the
messages emitted for file named `index.html` in some subdirectory are both
going to report the filename as just `index.html` even though they're
actually different files.


sideshowbarker:
Superseded by https://github.com/validator/validator/commit/8c4b1583883fa8656106aedac379de208e062f12

PR #102, opened 2015-06-08
inputmode attribute support for input and textarea tags

denofevil:
inputmode attribute is present in [HTML5 spec](http://www.w3.org/TR/html5/forms.html#the-input-element) without any details, [HTML5.1 editor draft](http://www.w3.org/html/wg/drafts/html/master/semantics.html#input-modalities:-the-inputmode-attribute) is giving better specification as well as [WHATWG living standard](https://html.spec.whatwg.org/multipage/forms.html#attr-fe-inputmode). Currently inputmode attribute is not validated properly by validator.
I don't know what policy is for those not-really-released features, but this PR can be useful when inputmode finally makes its way into spec.

### DIFF
--- a/schema/html5/web-forms.rnc
+++ b/schema/html5/web-forms.rnc
@@ -35,6 +35,23 @@ datatypes w = "http://whattf.org/datatype-draft"
 		attribute size {
 			common.data.integer.positive
 		}
+
+	shared-form.attrs.inputmode =
+		attribute inputmode {
+			w:string "verbatim" |
+			w:string "latin" |
+			w:string "latin-name" |
+			w:string "latin-prose" |
+			w:string "full-width-latin" |
+			w:string "kana" |
+			w:string "kana-name" |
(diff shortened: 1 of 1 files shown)


PyDeq:
BTW, according to this statement, the attribute is really valid: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27229#c5

PyDeq:
And IIUIC, this fixes https://www.w3.org/Bugs/Public/show_bug.cgi?id=23318.

sideshowbarker:
Thanks for putting together a patch for this. In light of the fact that some (most?) browsers still don’t support `inputmode` at all, I’m closing this in favor of just changing the error for `inputmode` into a warning, as landed in https://github.com/validator/validator/commit/3a062307f984a487f1abad59d1a63cade1ee4ebc.

I’ll revisit this if/when we ever get wider support for `inputmode` across more browsers.

PR #149, opened 2015-08-13
Fix "Ignore input[type=hidden] label descendants."

takenspc:
https://github.com/validator/validator/commit/d1d38e5905c1701d33ae7133a12b339e3aeced98
makes validator ignore input[hidden]. This commit makes validator not ignore input[hidden] but input[type=hidden].

Signed-off-by: Takeshi Kurosawa <taken.spc@gmail.com>


sideshowbarker:
oops… Thanks for catching this

PR #150, opened 2015-08-13
Use portable directives in `time.strftime`

takenspc:
This commit changes `validatorVersion` to "15.08.01" from "15.8.1"

References:

- https://docs.python.org/2/library/time.html#time.strftime
- http://strftime.org/

Signed-off-by: Takeshi Kurosawa <taken.spc@gmail.com>

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -62,7 +62,7 @@ nightliesPath = "/var/www/nightlies"
 releasesPath = "/var/www/releases"
 
 releaseDate = time.strftime('%d %B %Y')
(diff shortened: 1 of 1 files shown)


sideshowbarker:
Why?

The `15.8.1` formatting (without the zeroes) is intentional, not an oversight. What’s the benefit of having it padded out with zeroes? Does the lack of zeroes cause any actual problems?

sideshowbarker:
Closing, since as mentioned in https://github.com/validator/validator/pull/150#issuecomment-130528621 the single-padded-digits `16.1.1`-style formatting is intentional.

PR #218, opened 2016-01-11
Removing all unnecessary library dependencies for clients who use validator only as a java library

KristoficNetbeans:
I removed all unnecessary library dependencies and many of server related code for having minimal validator java library. I returned using IRI java library instead of galimatias for URL checking as it was in older validator versions. It is smaller java library.


sideshowbarker:
To share this with others I’d suggest you provide the source through your fork and for any maven-packaged version of it you might want to share, a separate group+artifact of your own.

I obviously can’t check this into the master branch here, and while it’s imaginable that we could check it into a separate branch here and distribute it through maven as a separate artifact, I don’t think it would make sense to do that by way of this repo—because I can’t commit any time to supporting it (by responding to any bug reports related to, or to reviewing any PRs against it that might come along) and also I’m not clear on the rationale for some of the details (for example, reverting to the Jena IRI library).

Anyway, if you distribute this yourself, there’s not requirement for me or anybody else to understand all the details, and you’ll have all the flexibility you want to evolve it further however you want.

But while I can’t accept this PR, of course feel free to post further comments here.

PR #219, opened 2016-01-13
build.py: Make it work with Python 3 too

scop:
Signed-off-by: Ville Skyttä <ville.skytta@iki.fi>


sideshowbarker:
This rocks. Thanks much for taking time to put it together.

PR #222, opened 2016-01-13
Tell Eclipse to use Java 8

scop:


sideshowbarker:
In https://github.com/validator/validator/commit/c60bd190a20b0d32f99cf4aefa7b13e0c73770e0 I pushed just the change to the ` .settings/org.eclipse.jdt.core.prefs` file.

If it turns out we need the `org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8` then we can handle that in a separate PR.

PR #243, opened 2016-02-21
Use try-with-resources more

scop:


sideshowbarker:
Thanks much—it’s very nice to be getting these kinds of code improvements made (I wish I were a more up-to-date and idiomatic Java programmer so that I could catch more of them myself too)

scop:
My Eclipse settings are set to flag a lot of things as errors, that way it's pretty easy to spot them :)

Let me know if you'd like me to submit a PR that adds a bunch of those settings.

sideshowbarker:
> Let me know if you'd like me to submit a PR that adds a bunch of those settings.

Yeah, that would be great (Eclipse is another thing I’m not very adept at—I don’t actually even use it directly but instead by way of [Eclim](http://eclim.org/).

PR #281, opened 2016-04-11
handle cookies in checkHttpURL()

valedelobo:
I need to test site that requires cookies saving. But now it impossible. I get error: `“java.net.ProtocolException: Server redirected too many times(20)”. `
With this changes, I can handle this error.
So, if you want, you can merge this commit into main repo.

Thanks.


sideshowbarker:
Thanks!

PR #287, opened 2016-05-01
Make it easier to update HTML spec copies

takenspc:
It seems that updating HTML spec copies is relatively complex. In particular,

- it requires a perl script to make HTML spec copies parsable
- it is not clear whether a given HTML spec copy is parsable

More straightforward way is 

- There should not be additional steps other than downloading
- Provide a way to check whether a given HTML spec copy is parsable

In this PR,

- added support "Tag omission in text/html" to support parsing raw HTML specs
- added a command to check whether a given HTML spec copies are parsable

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -102,7 +102,6 @@ log4jProps = 'resources/log4j.properties'
 heapSize = '128'
 stackSize = ''
 html5specLink = 'https://html.spec.whatwg.org/multipage/'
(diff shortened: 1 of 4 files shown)


sideshowbarker:
@takenspc Thanks extremely much for taking this on and making it much more sane. This will make it much easier to keep up to date going forward.

PR #316, opened 2016-07-07
Added new roles from ARIA 1.1 and DPUB-ARIA

halindrome:
These are new roles defined in ARIA 1.1 and in dpub-aria.


TzviyaSiegman:
This looks good to me (co-ed of DPUB-ARIA)

PR #336, opened 2016-08-19
Improve Markdown formatting of some commands

tripu:


sideshowbarker:
The `README.md` file is generated from the `index.html` file in the same directory. For details, [see the makefile](https://github.com/validator/validator/blob/master/site/Makefile#L12)

So I’m happy to merge the changes to `README.md` as long as we can generate them from changes either to the `index.html` source or changes to the makefile.

tripu:
After installing [`aaronsw/html2text`](https://github.com/aaronsw/html2text) and running the *makefile* I see spurious changes in `README.md`: my line length seems to be different, so some lines wrap into the next one at a different word even if I haven't changed them; linked URLs are inlined, instead of listed separately; etc. (Are these configuration options? Did I fail to set them up?)

Apart from that, I don't see how we can change the HTML so that blocks of code inside lists are rendered OK when transformed into Markdown; which was the purpose of this PR. (Examples: in [this HTML](https://github.com/validator/validator/blob/master/index.html#L447-L451) the command `java -cp …` [isn't in monospace](https://github.com/validator/validator/blob/master/README.md#http-client-for-fast-command-line-checking), this [other HTML](https://github.com/validator/validator/blob/master/index.html#L603-L606) produces [wrong Markdown](https://github.com/validator/validator/blob/master/README.md#build-instructions)).

So, closing this PR.

(I still wonder if it's worth keeping these two versions of the documentation; when I type `<command> --help` I don't expect to get nine screenfuls of Markdown on the console. I would prefer both the CLI and the `README.md` here on GitHub to be concise, and to point to [the online documentation](https://validator.github.io/validator/). Also, as it is now, we either make the HTML so simple that `html2text` can convert it OK, or we keep this wrongly-formatted Markdown.)

PR #376, opened 2016-09-28
Fix some markdown monospace issues

sideshowbarker:
Relates to https://github.com/validator/validator/pull/336

### DIFF
--- a/site/Makefile
+++ b/site/Makefile
@@ -23,4 +23,9 @@ all: ../README.md
 	    | $(PERL) $(PERLFLAGS) -pe 'undef $$/; s/(\n\n\n)+/\n\n/g' \
 	    | $(PERL) $(PERLFLAGS) -pe 's|(# The Nu Markup Checker \(v.Nu\))|$$1 $(BUILDSTATUS)|g' \
 	    | $(PERL) $(PERLFLAGS) -pe 'undef $$/; s/\s+(\[1\]:)\s+/\n   $$1 /' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|      java -cp|        java -cp|' \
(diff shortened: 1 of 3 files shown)


site/Makefile
```diff
@@ -23,4 +23,9 @@ all: ../README.md
 	    | $(PERL) $(PERLFLAGS) -pe 'undef $$/; s/(\n\n\n)+/\n\n/g' \
 	    | $(PERL) $(PERLFLAGS) -pe 's|(# The Nu Markup Checker \(v.Nu\))|$$1 $(BUILDSTATUS)|g' \
 	    | $(PERL) $(PERLFLAGS) -pe 'undef $$/; s/\s+(\[1\]:)\s+/\n   $$1 /' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|      java -cp|        java -cp|' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|    export|        export|' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|    git clone|        git clone|' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|    cd validator|        cd validator|' \
+	    | $(PERL) $(PERLFLAGS) -pe 's|    python|        python|' \
```
tripu: :bulb: Idea: replace all these with:

```bash
| $(PERL) $(PERLFLAGS) -pe 's|      $ |        $ |' \
```

and prepend `$ ` to every command in the original `index.html`, now and in the future.

sideshowbarker: Thanks, will try that out after merging this for now. There are other code-block places in the output that are indented 4 spaces but render fine (since they’re not nested under numbered steps or whatever) so want to make sure not to bork those.

PR #389, opened 2016-10-21
Update to salvation 2.2.0 (fix .classpath reference)

leonatherton:
Fixes broken reference to old salvation version


sideshowbarker:
Thanks for catching that 🖖

PR #442, opened 2017-01-14
Replace colons in links to the wiki with "»"

xfq:
Signed-off-by: Xue Fuqiao


sideshowbarker:
Thanks for catching this

PR #452, opened 2017-01-23
New EmbeddedValidator

mdewilde:
Add a way to use the Nu HTML validator as library in another Java application.


sideshowbarker:
Thanks very much for contributing this—it’s great to have this available and I’m certain there are others who will find it really useful.

Issue #154, opened 2015-09-07
Unchecking 'Show Source' still shows source in Text Input mode


BoffinbraiN:
To reproduce, switch to Text Input mode and paste a document to validate. Uncheck the 'Show Source' checkbox and then validate. The source will be shown at the bottom of the page anyway.

sideshowbarker:
Actually this is expected behavior. The reason is that we copy the source back into the text-input box in the results page. And we do that in the frontend code, on the client side. And so we need to copy it from somewhere, which is, from that source view.

I can probably (maybe) add some JS code to be a bit more clever about, in the "Show source"-unchecked case, deleting the source from the DOM after we copy it, or just hiding it with CSS. But it’s not a super high priority and it could be a while before I get to it, so in the mean time, patches welcome.

BoffinbraiN:
Oh... Couldn't the server-side code return the page with the HTML already inside the textarea, and then choose to not emit the source view? I had a very quick look at the project but couldn't find the code for the controller - just the XML that makes the front-end page.

sideshowbarker:
Reverted https://github.com/validator/validator/commit/43e8f992d5adc62c6e2a4bbf65734f14e344da1a

PR #479, opened 2017-03-20
Fix the behavior of "Show Source" in text input mode

xfq:
See #154.


sideshowbarker:
Beautiful—thanks much, glad to have this one fixed finally

PR #487, opened 2017-03-28
(WIP) Support the CSS <color> production

### LINKED ISSUE #483: Add datatype/microsyntax checking of color values for link[rel=mask-icon] and meta[theme-color]
Now that we allow `link[rel=mask-icon]` and `meta[theme-color]`, which both enable CSS color values to be specified, we should do actual datatype/microsyntax checking on the values of the accompanying attributes for those which the HTML spec says must be CSS color values.

https://github.com/validator/validator/blob/master/src/nu/validator/datatype/Color.java is stub where the actual datatype checking would need to go.

Implementing the datatype checking could be a lot of work, given the ridiculously wide range of ways to specify colors in CSS https://drafts.csswg.org/css-color/#typedef-color and also given baroqueness syntax allowed with that.

https://github.com/w3c/css-validator/tree/master/org/w3c/css/values has some code we might be able to steal from; see in particular:

* https://github.com/w3c/css-validator/blob/master/org/w3c/css/values/CssColor.java
* https://github.com/w3c/css-validator/blob/master/org/w3c/css/values/CssColorCSS4.java

xfq:
See #483.

The patch is far from complete, but @sideshowbarker do you have any comment on it? The earlier the problems are found and fixed, the less expensive the fixes are.

### DIFF
--- a/src/nu/validator/datatype/AbstractDatetime.java
+++ b/src/nu/validator/datatype/AbstractDatetime.java
@@ -36,7 +36,7 @@ import org.relaxng.datatype.DatatypeException;
  */
 abstract class AbstractDatetime extends AbstractDatatype {
     /**
-     * Days in monts on non-leap years.
+     * Days in months on non-leap years.
      */
     private static int[] DAYS_IN_MONTHS = { 31, 28, 31, 30, 31, 30, 31, 31, 30,
             31, 30, 31 };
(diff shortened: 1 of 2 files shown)


xfq:
Closing for now.

Will create a new PR after more exploration.

PR #512, opened 2017-05-17
Follow-up for the multiple main elements PR

xfq:
Fix a bug introduced in #496. Before this fix, the error will not be emitted whatever the circumstance.


xfq:
If there is interest, here are two diagrams about this fix (sorry, they are not in SVG):

img_20170517_211152
img_20170517_211212

PR #513, opened 2017-05-18
Update CSS files.

XhmikosR:
* use shorthands when possible
* no need to use IDs with elements
* remove units from zero
* use lowercase color hex values consistently
* add the unprefixed `transition` property

### DIFF
--- a/site/nu-style.css
+++ b/site/nu-style.css
@@ -2,26 +2,26 @@
  * Copyright (c) 2005, 2006, 2007 Henri Sivonen
  * Copyright (c) 2007-2014 Mozilla Foundation
  *
- * Permission is hereby granted, free of charge, to any person obtaining a 
- * copy of this software and associated documentation files (the "Software"), 
- * to deal in the Software without restriction, including without limitation 
- * the rights to use, copy, modify, merge, publish, distribute, sublicense, 
- * and/or sell copies of the Software, and to permit persons to whom the 
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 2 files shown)


site/nu-style.css
```diff
@@ -860,7 +847,7 @@ dl.inputattrs .highlight {
 }
 
 #filters code {
-	background: #6F6;
+	background-color: #6f6;
```
xfq: There are two `background-color` declarations in this rule.

site/nu-style.css
```diff
@@ -70,38 +70,38 @@ textarea {
...
 }
 
-h1#title {
+#title {
 	font-family: "Myriad Web", "Myriad Pro", "Gill Sans", Helvetica, Arial, Sans-Serif;
-	background-color: #365D95;
+	background-color: #365d95;
 	color: #fdfdfd;
 	font-size: 1.6em;
 	padding: 0.430em;
```
xfq: s/430/43/

site/style.css
```diff
@@ -67,7 +67,7 @@ h3 {
 }
 
 h1 span {
-	color: #AAAAAA;
+	color: #aaaaaa;
 }
 
 fieldeset td, fieldeset th {
```
xfq: s/fieldeset/fieldset/

XhmikosR: No need to comment on unrelated stuff to my changes.

xfq: I apologize for my laziness. I should create a separate PR for this.

XhmikosR: Yeah, exactly my point :P

Nevertheless, I addressed your previous comments in this branch, but for any others just make a PR.

PR #563, opened 2017-09-22
Fix leak in LanguageDetectingXMLReaderWrapper.initialize()

tgyurci:
LanguageDetectingXMLReaderWrapper.initialize() function does not clean
the languageTags and profiles static lists.  Therefore calling
initialize() again doubles the content of these lists.  This can lead to
OutOfMemoryError.  Making languageTags and profiles local can fix this.


sideshowbarker:
Thanks much!

PR #691, opened 2018-08-25
[CHANGELOG.md] Correct element for which 'allow' attr is allowed.

Malvoz:
The `allow` attribute is allowed for `<iframe>` and not ``, was introduced in https://github.com/validator/validator/pull/686


sideshowbarker:
Thanks much〜

PR #712, opened 2018-10-05
Build: exit 1 if subprocess exit status non-zero

sideshowbarker:
This change makes the build exit with 1 as the status code if any subprocess.call has a non-zero exit status. That will cause CI to see the right exit code so that, e.g., travis_retry will work as expected in cases where, e.g., the css-validator (sub)build fails to download one of its dependency jars.


XhmikosR:
It seems it breaks AppVeyor completely. I pushed a patch to use `appveyor-retry`, not sure if this will fix it...

XhmikosR:
Must be those 2 files don't have end of file newlines (tests/html/elements/input/pattern-asterisk-novalid.html and tests/html/elements/input/pattern-paren-novalid.html)

I don't get why AppVeyor doesn't show the errors though and we need to get the Raw log to see them. Reported in https://github.com/appveyor/ci/issues/2682

XhmikosR:
All good with retry. AppVeyor worked as expected since cache it didn't have the cache and retried 3 times and then failed. Travis without cache retried too.

So I think we are good to go.

PR #730, opened 2018-11-26
Update .travis.yml

XhmikosR:
Remove `sudo: false` and default `dist`.

https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


XhmikosR:
Let's put this on hold since build breaks without `sudo: false`.

PR #743, opened 2019-01-16
add focusin and focusout events

denofevil:
Added `onfocusin` and `onfocusout` attributes from https://www.w3.org/TR/uievents/


sideshowbarker:
Thanks!

PR #757, opened 2019-02-07
Allow ARIA role=figure on figure element

rdeltour:
`role=figure` is allowed on `figure` elements (even if a SHOULD NOT) in [ARIA in HTML](https://w3c.github.io/html-aria/#figure).

Currently, the following snippet:

```html
<figure role=figure>figure</figure>
```

is reported both as an error and a warning. This change should make the error go away.


sideshowbarker:
Thanks much

PR #786, opened 2019-04-21
Test on JDK 12 and remove 10 from AppVeyor.

XhmikosR:
@sideshowbarker: not sure what's the policy for the supported Java versions, I feel by keep adding more JDK versions will slow CI a lot, but personally I think it'd be good to test the latest version too.

Let me know how you want to proceed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,8 +7,8 @@ environment:
   BUILD_PY: .\build\build.py
   matrix:
     - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
-    - JAVA_HOME: C:\Program Files\Java\jdk10
     - JAVA_HOME: C:\Program Files\Java\jdk11
+    - JAVA_HOME: C:\Program Files\Java\jdk12
 
 install:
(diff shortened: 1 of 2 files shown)


sideshowbarker:
> @sideshowbarker: not sure what's the policy for the supported Java versions,

I see now that https://www.oracle.com/technetwork/java/java-se-support-roadmap.html and other places make it clear that Java9 no longer has any official support — it’s not getting any updates, not even bug fixes — and Java10 is not far from that point.

The gist of it is that Java8 and Java11 still have official support, so I think we should still be testing with them. And Java12 was released last month, so I think we should be testing with it now too.

I will be removing Java9 testing — because it is broken now anyway by a recent change I made that should not actually be breaking and that I don’t want to waste any more time troubleshooting.

I can leave Java10 for now — because it’s not broken anyway. But personally I think we should just remove Java10 along with Java9.

> I feel by keep adding more JDK versions will slow CI a lot, but personally I think it'd be good to test the latest version too.

Yes we should be testing with whatever the latest version is. But with Java9 and Java10 both removed, we’d reduce to just three versions: 8, 11, and 12.

PR #790, opened 2019-04-23
AppVeyor: remove `JAVA_HOME\bin` from `PATH`

XhmikosR:
It's not needed after 9fa8c3c

BTW @sideshowbarker perhaps we should use `os.path.join` in this?

https://github.com/validator/validator/blob/5695696bfacaf83c39105ae46f2b23d39cf33f94/build/build.py#L62-L67


XhmikosR:
Hmm, I might need to change `java -version` later or just drop this patch, I'm not sure.

Issue #187, opened 2015-12-02
Add support for "warning" level


estan:
It would be nice to be able to see errors and warnings, but not informational messages.

Umoxfo:
I'm working on it, and I made pull request #819.

PR #819, opened 2019-05-23
Add support for "warning" level

Umoxfo:
#187

### DIFF
--- a/src/nu/validator/messages/types/Warning.java
+++ b/src/nu/validator/messages/types/Warning.java
@@ -22,7 +22,7 @@
 
 package nu.validator.messages.types;
 
-class Warning extends Info {
+class Warning extends MessageType {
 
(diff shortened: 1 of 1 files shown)


sideshowbarker:
I’m going ahead and closing this, because for one thing, you don’t explain at all what exactly this change actually does.

The feature request in https://github.com/validator/validator/issues/187 was this:

> It would be nice to be able to see errors and warnings, but not informational messages.

It’s unclear to me if this patch even does that. But even if it does, it seems clear that as-is in its current form, that change could only be doing causing informational messages to never be reported any longer.

If so, then we can’t do that. What would instead be needed is some way for the user to choose to _optionally_ suppress all info-level messages — and only see warnings and errors.

That said, I’m not sure I would merge such a allow-users-to-suppress-all-info-messages patch even if somebody were to raise a PR for it.

PR #884, opened 2019-11-01
Assertions: Sort attribute lists at construction to fix race condition

andersk:
`Arrays.sort` is not thread-safe, so concurrently validating many documents with `<input type="text" placeholder="foo"/>` would occasionally result in confusing errors like `error: Attribute “placeholder” is only allowed when the input type is “email”, “number”, “password”, “search”, “search”, “tel”, or “url”`.

This may explain a bug report posted three years ago: http://lists.w3.org/Archives/Public/www-validator/2016Dec/0013.html


andersk:
When you get a chance, can you update https://www.npmjs.com/package/vnu-jar with this fix?

sideshowbarker:
https://www.npmjs.com/package/vnu-jar/v/20.1.2 published

PR #912, opened 2020-02-29
Travis variable cleanup

nschonni:
Was reading through to see if I could figure out a cron for the Docker and found a few of the indirect variables a little hard to follow.
- Pulls up all the in-line `export` to global or matrix variables
- Pulls Tomcat setup into the `install` step

I think you have some logic encrypted in `resources/ssh_travis.enc` so maybe I'm missing part of the need for some of the variables


sideshowbarker:
Reviewed and tested and it merged to master and it all works as expected — much thanks!

PR #913, opened 2020-02-29
chore: Merge to single GitHub release for publish

nschonni:
Creates a single "develop" pre-release tag and upload all assets to tag.

- This will require updated the Dockerfile since it's using the platform specific tag. I'd probably do a second pass after this to ensure the version number tags also get populated with these assets, and use those as the Dockerfile assets.

- Appveyor would also needed to be updated to include the Windows assets.

Maybe you ran into an issue with this type of setup originally, so I'm fine if you don't like this approach and just want to close it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ git:
 
 branches:
   except:
-    - /^(osx|linux|windows|jar|war)$/
+    - /^(osx|linux|windows|jar|war|develop)$/
 
 env:
   global:
@@ -94,11 +94,12 @@ before_deploy:
   - ls -al build/dist-war
   - git config --local user.name "Travis CI"
   - git config --local user.email "mike@w3.org"
-  - git push --delete git@github.com:validator/validator.git $TRAVIS_TAG || true
-  - git tag -m $TRAVIS_TAG -f $TRAVIS_TAG -s --local-user 0x6C976B92
+  - git push --delete git@github.com:validator/validator.git develop || true
+  - git tag -m develop -f develop -s --local-user 0x6C976B92
 
 deploy:
   - provider: releases
+    tag_name: develop
     overwrite: true
     prerelease: true
     file_glob: true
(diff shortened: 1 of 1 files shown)


sideshowbarker:
I think the effect of having only one pre-release tag is that when users go to https://github.com/validator/validator/releases they wouldn’t see separate "windows", "war", "osx", "linux", "jar" releases, but instead would see only one "develop" release.

And I think that one "develop" release would have only one set of assets — the assets for whichever release target/platform was built last. Or even if not that, it would have all of the assets for all targets/platforms together — which would be a long list they'd need to hunt through just to find the one set that they actually wanted.

Anyway, the whole way I have Travis configured to upload and tag the pre-releases is a hack — because I couldn’t find anything other way to do is that would work.

Maybe migrating over to GitHub Actions would give us a better way to do it.

nschonni:
I'll take a look to see what it would look like in Actions-land. I'm guessing you're thinking about https://github.com/actions/upload-artifact/ which would just have the binaries attached to each build.

sideshowbarker:
> I'll take a look to see what it would look like in Actions-land.

That’d be really great

> I'm guessing you're thinking about https://github.com/actions/upload-artifact/ which would just have the binaries attached to each build.

Yeah, definitely that specifically — but also, overall, I have a vague sense that migrating to GitHub Actions and away completely from Travis for all of our current CI would be better for the long term, as far as ease-of-maintenance and getting around Travis limitations.

But as far as the particular case of handling the artifacts for pre-releases, it seems like the limitation might be in in the GitHub UI itself — in that, in order for some set of artifacts/assets to be listed separately at https://github.com/validator/validator/releases, it needs to have a separate tag.

It would be great if there were instead a way in the GitHub release UI to have all of the releases under one listing on https://github.com/validator/validator/releases, but with a separate sub-list/title for each platform/target — a `<dialog> for each, with its own **▶ Assets** summary. 

Otherwise, lacking that, the only way I can see is that all the assets will get tossed into a single long list, with only one **▶ Assets** \<dialog> for the whole of them (40 items).

nschonni:
If you want to try out this approach will the GitHub actions gets tested out, I'd suggest just adding some static Markdown to the release description to highlight the main assets like: 
```md
## Linux
- [Download jar](https://github.com/validator/validator/releases/download/develop/vnu.linux.zip)

...
## OSX
- [Download jar](https://github.com/validator/validator/releases/download/develop/vnu.osx.zip)
...
````

sideshowbarker:
> If you want to try out this approach will the GitHub actions gets tested out, I'd suggest just adding some static Markdown to the release description to highlight the main assets

OK, yeah that could work

sideshowbarker:
This was fixed when we moved to GitHub Actions

Issue #914, opened 2020-02-29
Convert to GitHub Actions for CI


nschonni:
Pulling out the discussion from #913 to make it easier to track.

- [ ] matrix of Linux/OSX JDK builds to match current travis on push/PR
- [ ] matrix of Windows JDK builds to match current appveyor on push/PR
- [ ] upload artifacts on push to master to replace osx/linux/windows/jar/war tags
- [ ] job upload artifacts on final git tags
- [ ] job to publish Docker hub with version number
- [ ] job to publish npm on final tag

nschonni:
@sideshowbarker is there a reason some of the jobs use Python 3 vs some not?

sideshowbarker:
The reason for some of the jobs using Python 3 vs some not is just to ensure we are testing both with Python 2 and Python 3. So the association of any particular job with Python 2 is just arbitrary. We only need to test one job with Python 2, but it doesn’t matter which job it is.

nschonni:
OK, I don't think it would be hard to set up a python check/lint run separate from the regular build, if you know what it should cover

sideshowbarker:
The Python 2 check should make sure that all the build targets we call in CI work under Python 2. The goal is to ensure that when we’re making changes to the `build/build.py` script, we’re not introducing regressions that prevent the script from continuing to work under Python 2.

sideshowbarker:
This is done now

PR #918, opened 2020-03-01
Initial GitHub Actions setup

nschonni:
This is pretty messy for the commits because it's not passing everywhere. Current state: 

- [x] Linux builds with Artifact uploads
- [ ] OSX build errors
- [x] Windows passes on 8, fails the rest
- [ ] There is a bunch of duplication for OSX/Linux that should get pulled out to separate YAML files to share
- [ ] Releases aren't signed
- [ ] Files are double zipped do to an issue with v1 of actions/upload-artifact that may be resolved with v2 release
- [ ] Only uses Python 3. Python 2 could be a separate job with PyLint or others though
- [ ] Rebase commits before landing

Related to #914


.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+        name: vnu.windows.zip
+        path: "build/dist/vnu.windows.zip"
+
+  darwin:
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        java:
+        - 11
```
sideshowbarker:
```suggestion
```

For our purposes, for OSX, we don’t need to specify any particular Java version. Whatever Java version it defaults to on OSX is fine.

nschonni: I'll check to see if there is one, but I'm not sure if OSX has a default JDK, or just a JRE version

nschonni:
Also meant to share that this is what it looks like right now https://github.com/nschonni/validator/actions/runs/47735481
I'm not sure if the OSX failure is related to the slight version difference since it is slightly newer and uses Zulu instead AdoptJDK as the source

sideshowbarker:
Nick,

I think if we have even just the core build working under CI in GitHub Actions — and even if it’s only under Linux — that’s sufficient and we should go ahead and merge the working parts of this — including ripping out all the Travis/AppVeyor stuff.

---

Longer comments:

First off, thanks extremely much for having put this all together. I should have gotten around to following up on this a long time ago, but anyway I’m hoping we can do a new major release sometime soon, and would like if we can finally completely drop the existing Travis CI and AppVeyor CI and replace it with this.

So if you still have time and interest in getting this landed, that would be super great.

> This is pretty messy for the commits because it's not passing everywhere. Current state:
> 
> * [x]  Linux builds with Artifact uploads

As I said above, I think having the core build working under CI — even if for now it’s just Linux only — we should consider that sufficient for getting this merged.

> * [ ]  OSX build errors
> 
> * [ ]  Windows passes on 8, fails the rest

I propose we deal with the CI for the core build under OSX and Windows later, over the longer term.

And as far as the parts of the build that create the OSX and Windows binary images — that I think can definitely be just an extremely low priority, even over the long term.

As far as having the “normal” core CI for the non-release jar and war running under OSX and Windows, it’s of course less than ideal to stop running those for now and so not have any idea if some change broke those.

But I think for now it’s more important to just have working GitHub Actions CI for the core build running under *any* environment at least — as a start. So we can land that and deal later with getting CI for OSX and Windows working again.

And as far as the builds for the binary images, in general I have no idea how many people might actually be using the binary image releases other than the Linux image (which is used for building the Docker image and so which I know actually works).

And I hate to risk breaking things for people but at this point if we don’t even have feedback from people using the OSX and Linux binary images, we also don’t have a good way to know when we’ve broken them in some subtle or surprising way that we can’t catch with just CI.

So IMHO trying to get the OSX and Windows binary-image builds working under GitHub Actions is espcially not worth the cost to us in the near term — and even more particularly not worth the cost if it’s the only thing blocking us from migrating to GitHub Actions and away from Travis and AppVeyor.

If there are people who are actually relying on the releases of the OSX and Windows images, then after they become available, we might get some actual issues/feedback from them, and then we’d have some evidence to use for deciding how to prioritize the need.

> * [ ]  There is a bunch of duplication for OSX/Linux that should get pulled out to separate YAML files to share

I think we can deal with that later if/when we can get the OSX build working again.
 
> * [ ]  Releases aren't signed

I think we can live without that for now.

> * [ ]  Files are double zipped do to an issue with v1 of actions/upload-artifact that may be resolved with v2 release

I don’t know if there’s been a v2 release yet but regardless I don’t think we need to let the double zipping be a show stopper. We can document it, and deal with any issues that get reported about it from users who we’ve confused.

> * [ ]  Only uses Python 3. Python 2 could be a separate job with PyLint or others though

I think at this point we can abandon the Python 2 build — and even abandon Python 2 compatibility. Would you agree?

As far as I understand, most or all OS distros have now moved to Python 2 — so if there are users who are still relying on Python 2 support, I’d wonder who they are and why.

But without feedback from anybody actually relying on Python 2 support, I think continuing to try to ensure Python 2 compatibility is not worth the ongoing cost.

---

Anyway, overall just looking forward to hearing your thoughts.

sideshowbarker (approved):

sideshowbarker:
Should we go ahead and merge this?

nschonni:
Was just going to try your suggestion of trying the default Java for macOS to see if it makes it happy. Otherwise I think you can squash and merge if you just want to handle that after

sideshowbarker:
> Was just going to try your suggestion of trying the default Java for macOS to see if it makes it happy. Otherwise I think you can squash and merge if you just want to handle that after

Well I see you pushed the default Java change. I guess let’s see if that fixes anything. But if not, yeah, we can go ahead and merge and deal with that later.

sideshowbarker:
hmm OK I see https://github.com/nschonni/validator/runs/2922874804#step:3:11 says *Error: Input required and not supplied: java-version* so I guess the answers the question about whether the java version is needed or not.

sideshowbarker:
For now shall we just merge the Linux yaml and Windows yaml (now that you got that fully working)? (I mean along with removing the Travis and AppVeyor stuff)

nschonni:
I just reverted that commit and re-pushed if you just want to squash/land as-is for now

nschonni:
Oh, do you want me to just drop the macOS build entirely for now?

sideshowbarker:
> Oh, do you want me to just drop the macOS build entirely for now?

Well I was just running l thinking that if the OSX part is gonna be preventing the CI on the main branch from being green, then we maybe we don't want to merge it but instead keep it in its own branch where we can keep working on it.

nschonni:
OK, I commented out the most, so it should "pass", but really it just checks out the code for now on macOS :p

nschonni:
EX: https://github.com/nschonni/validator/actions/runs/975207291

sideshowbarker:
> OK, I commented out the most, so it should "pass", but really it just checks out the code for now on macOS :p
>
> EX: [nschonni/validator/actions/runs/975207291](https://github.com/nschonni/validator/actions/runs/975207291)

Excellent — so you did actually make us green across all platforms 😆

Let’s declare victory and I can spend some time helping fight the OSX battle in the background

sideshowbarker:
@nschonni I took the liberty of setting you to admin access for the repo now — which I can imagine might be necessary for troubleshooting some types of build issues. But if you’d prefer not to have admin access, that’s fine too (and I can either flip it off, or I guess you should also be able to do so yourself)

nschonni:
No worries, I'll probably still submit PRs from my fork, but probably useful to have in case something hits a snag

sideshowbarker:
FYI the reason for the macOS build error was this:

https://github.com/validator/validator/blob/ed6a2fd37b2608be99a405679e57bb87962bdc0b/.github/workflows/darwin.yml#L45-L46

The output of `file` in a macOS environment doesn’t include what that assumes — specifically, it doesn’t include and lines with `version` at all.

But GitHub Actions didn’t make it easy to figure that out. It took a lot of trial and error to isolate that as the problem. In hindsight maybe I should have tried the thing of adding the `ACTIONS_STEP_DEBUG` secret to turn on debug logging.

Anyway, I’ve fixed it locally now — and also added steps for uploading releases in the same was we had that working previously with Travis and AppVeyor.

So sometime later today, I’ll open a PR with the changes for review.

Issue #915, opened 2020-03-01
Validation for aria-dropeffect fails valid combinations


dd8:
https://www.w3.org/TR/wai-aria-1.1/#aria-dropeffect

I can't see anything in the normative text that prohibits certain combinations of drag-and-drop operations. In particular the copy+move combination is flagged as invalid, but is very common in file drag-and-drops:

```
aria-dropeffect='copy move'
``` 

aria-dropeffect is marked as deprecated in ARIA 1.1 (and probably has close to zero support) so simplest fix might be removing validation of aria-dropeffect values if this is tricky to express in RNC

sideshowbarker:
> aria-dropeffect is marked as deprecated in ARIA 1.1 (and probably has close to zero support) so simplest fix might be removing validation of aria-dropeffect values if this is tricky to express in RNC

Yeah, given that I went ahead and changed it to allow any value

dd8:
Sorry, discovered  some ACT Rules test cases for both valid and invalid aria-dropeffect values:
https://act-rules.github.io/rules/6a7281#passed-example-9
https://act-rules.github.io/rules/6a7281#failed-example-8

I submitted a PR with an updated fix

PR #920, opened 2020-03-03
Updated aria-dropeffect fix

dd8:
This is an update to the fix I suggested in issue https://github.com/validator/validator/issues/915

After I reported the issue I discovered some ACT Rules test cases for both valid and invalid `aria-dropeffect` values:
https://act-rules.github.io/rules/6a7281#passed-example-9
https://act-rules.github.io/rules/6a7281#failed-example-8

This patch should now correctly check invalid and valid values per ARIA 1.1


sideshowbarker:
Thanks!

PR #967, opened 2020-05-22
Url list option

JustMisha:
adding an option --url_list FILENAME to check files (urls) specified in file FILENAME where each line of the file contains a file (URL). If the option is set checker validates both the files in this file and the ones specified in command-line arguments. That allows start vnu.jar only once to check any numbers of pages.

Also adding in cli-help:
#### --url_list _FILENAME_

    Specifies a filename with files (URLs) to be checked. Each line of the 
    file contains a file (URL). If the option is set checker validates both 
    the files in this file and the ones specified in command-line arguments.

    default: [unset; vnu validates only files (URLs) specified after all the 
    options or in standard input]

### DIFF
--- a/src/nu/validator/client/SimpleCommandLineValidator.java
+++ b/src/nu/validator/client/SimpleCommandLineValidator.java
@@ -269,7 +269,7 @@ public class SimpleCommandLineValidator {
                                 + " requires a URL for a schema.");
                         System.exit(1);
                     }
-				} else if ("--file_url_list".equals(args[i])) {
+				} else if ("--file-url-list".equals(args[i])) {
                     fileNameWithFileList = args[++i];
 					hasFileList = true;
                 }					


CSUJAY (approved):
The coding is good

sideshowbarker:
I am going ahead and closing this, though somewhat reluctantly — because I can see that the feature is coded up quite nicely in this patch. However, it’s not a feature that, on balance, brings enough benefit to justify the costs it brings in terms of adding further code to maintain, and yet more “options surface” for users to have to wade through.

And thing is, there are already ways that users can maintain lists of URLs and/or filenames in a file, and to have that list of URLs and/or files fed to the checker without the checker needed to be re-launched each time for each URL or file.

What I mean is, there are existing mechanisms and tools that can be used for it which are already available in the shell environments where users are running the command-line checker.

And given that, this change seems more at the level of being a convenience feature. And while convenience features are not bad things in general, I think this particular use case is not one that merits adding such a feature to this checker.

PR #987, opened 2020-06-22
allow doc-glossary on section

mattgarrish:
Per the [ARIA in HTML doc](https://www.w3.org/TR/html-aria/#el-section), doc-glossary should be allowed on section.


mattgarrish:
Sorry, just realized this is a duplicate. Was reported in epubcheck.

PR #988, opened 2020-06-23
add doc-dedication to aside roles

mattgarrish:
Found one more omission from ARIA in HTML: doc-dedication is [allowed on aside](https://www.w3.org/TR/html-aria/#el-aside).

I set up tests for all the elements and their allowed dpub-aria roles and this was the only new error I found, so everything should be good now.


sideshowbarker:
Thanks!

Issue #1008, opened 2020-08-03
Allow rel attribute on a element in svg 


mattgarrish:
The `rel` attribute is supported on `a` in SVG2 but isn't currently valid to use.

I found mention that full validation of SVG2 isn't in the cards, but the need for this particular attribute has been raised in https://github.com/w3c/epubcheck/issues/1172

Would be helpful if it could be added.

PR #1009, opened 2020-08-04
allow rel on svg a element

mattgarrish:
Updates the schema to allow a `rel` attribute on `a` elements as requested in #1008.

I took a look at the java code for validating the tokens, but as far as I can tell (being no java developer), the existing processing of `a` attributes should cover this once allowed. If I've got that wrong, just let me know where I should be looking to enable it.


sideshowbarker:
Thanks much, and sorry I didn’t get this merged before now

PR #1010, opened 2020-08-07
Fix an infinite recursion.

shlomif:
Hi!

checkJar called createJarOrWar which called checkJar.

Please look into reviewing and merging it.


shlomif:
@sideshowbarker : thanks for merging - and you're welcome. That was very quick too.

PR #1077, opened 2021-02-01
Update npm.md

XhmikosR:


sideshowbarker (approved):

XhmikosR:
Might be worth looking into switching to GitHub Actions later since Travis is a lot slower and recently they added some limitations.

nschonni:
@XhmikosR I started looking it it in https://github.com/validator/validator/pull/918, but haven't picked it back up

sideshowbarker (approved):

sideshowbarker:
> @XhmikosR I started looking it it in #918, but haven't picked it back up

Yeah I guess we really should move to GitHub Actions workflow stuff. I think what’s blocking it is just the idiosyncratic way I’ve got the release builds set up. Probably I should really look into doing that differently. But practically speaking, I don’t think I will manage to make time for it soon — because it will require a significant amount of time, and right now most of the time I’d have is taken up by mdn/content and mdn/browser-compat-data stuff (on top of more responsibilities I’ve taken on this year for W3C-specific work)

Issue #1118, opened 2021-03-14
updates needed for use of aria-required in HTML


scottaohara:
[test cases](https://w3c.github.io/html-aria/tests/required-test.html)

test 1 expects no warnings for use of `aria-required` on elements where the `required` attribute is allowed. presently `date`, `datetime-local`, `month`, `time` and `week` `input`s do not pass the validator.

test 2 expects a warning when `aria-required=true` and `required` are used on the same element. presently only `textarea` and `select` throw a warning for the duplicate use of this attribute.

test 3 expects an error when `aria-required=false` and `required` are used on the same element. presently this is not implemented

sideshowbarker:
> [test cases](https://w3c.github.io/html-aria/tests/required-test.html)
> 
> test 1 expects no warnings for use of `aria-required` on elements where the `required` attribute is allowed. presently `date`, `datetime-local`, `month`, `time` and `week` `input`s do not pass the validator.

Fixed by https://github.com/validator/validator/commit/78b8018c

> test 2 expects a warning when `aria-required=true` and `required` are used on the same element. presently only `textarea` and `select` throw a warning for the duplicate use of this attribute.

Fixed by https://github.com/validator/validator/commit/726e759

> test 3 expects an error when `aria-required=false` and `required` are used on the same element. presently this is not implemented

Fixed by https://github.com/validator/validator/commit/726e759

So I believe all problems reported in this issue are now fixed in the sources — and pushed to https://validator.w3.org/nu/ — and so I’m going ahead and closing this. But if I’ve missed anything, please let me know.

scottaohara:
thank you!

PR #1123, opened 2021-03-22
ARIA: Allow aria-required with special input types

revoltek-daniel:
Add aria-required to input types datetime-local / date / month / time / week 
Like proposed in: https://github.com/validator/validator/issues/1118


sideshowbarker (approved):
Looks great — thanks much

PR #1146, opened 2021-05-11
docs: switch to higher quality reference for alternative text

j9t:
(and remove unneeded whitespace)

Signed-off-by: Jens Oliver Meiert <jens@meiert.com>

### DIFF
--- a/schema/html5/assertions.sch
+++ b/schema/html5/assertions.sch
@@ -65,7 +65,7 @@ THE SOFTWARE.
 				&#x201C;alt&#x201D; attribute, except under certain
 				conditions. For details, consult guidance on
 				providing text alternatives for images.
-				http://www.w3.org/wiki/HTML/Usage/TextAlternatives
+				https://html.spec.whatwg.org/multipage/images.html#alt
 			</assert>
 		</rule>
 	</pattern>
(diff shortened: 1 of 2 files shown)


sideshowbarker (approved):

stevefaulkner:
@sideshowbarker while I agree that the current reference could be improved, saying that the https://html.spec.whatwg.org/multipage/images.html#alt is of higher quality just ain't so. The reason I spent years writing and updating https://www.w3.org/TR/html51/semantics-embedded-content.html#alt-text was because what was in the whatwg spec was not fit for purpose and whatwg content has been largely unmaintained for a decade. I suggest instead that https://www.w3.org/WAI/tutorials/images/ be used as a reference

j9t:
Just in case, this PR wasn’t meant to step on anyone’s toes. Sorry for not providing more detail on the PR, either. I actually believe the MDN reference is pretty good, I also think that it makes sense to refer to a URL on w3.org, even if it redirects, yet the HTML spec’s chapter on images seems not only “canonical,” but also really, really good. And their high quality led me to suggest this change.

Hope this offers additional perspective, without suggesting there was flaw with anyone’s work.

stevefaulkner:
@j9t no toes stepped on, thanks for bringing this up the HTML alt advice is poorly maintained and is not really fit for purpose as it does not align in places with the advice on appropriate alt for accessibility purposes, the one I pointed to https://www.w3.org/WAI/tutorials/images/ is maintained and fit for purpose

PR #1168, opened 2021-06-27
chore: Add Dependabot for GitHub Actions and submodules

nschonni:
The actions are mostly pinned to the major versions, but should give PRs for any new majors. Wasn't sure about the submodules, but it will create PRs for when those repos are changed as well


nschonni:
Could also add branch filters so in the GitHub Actions so it doesn't have the current "double" build for the PR + Push like right now

sideshowbarker (approved):

PR #1176, opened 2021-07-05
CI: Refactor macOS/Linux builds; add release automation

sideshowbarker:
This includes the following changes:

* factor out the common part of the macOS and Linux builds into a separate `.github/actions/build/action.yml` file

* add automation for releasing the jar and war files, as well as the Linux, macOS, and Windows binary runtime images

* fix a bug that was breaking the macOS build

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -1,0 +1,28 @@
+name: 'Build'
+runs:
+  using: "composite"
+  steps:
+    - id: build
+      env:
+        TOMCAT_VERSION: 8.5.29
+      shell: bash
+      run: |
+        python ./checker.py build
+        python ./checker.py jar
+        python ./checker.py image
+        python ./checker.py war
+        java -jar ./build/dist/vnu.jar ./build/dist/index.html
+        java -jar ./build/dist/vnu.jar ./site/nu-about.html
+        jar xf ./build/dist/vnu.jar && javap -verbose nu/validator/client/SimpleCommandLineValidator.class | grep "major version: 52"
+        curl -O https://archive.apache.org/dist/tomcat/tomcat-8/v$TOMCAT_VERSION/bin/apache-tomcat-$TOMCAT_VERSION.zip
+        unzip apache-tomcat-$TOMCAT_VERSION.zip
+        chmod a+x apache-tomcat-$TOMCAT_VERSION/bin/*.sh
+        jar xf ./build/dist-war/vnu.war && javap -verbose WEB-INF/classes/nu/validator/servlet/VerifierServlet.class | grep "major version: 52"
+        apache-tomcat-$TOMCAT_VERSION/bin/catalina.sh start
+        cp ./build/dist-war/vnu.war apache-tomcat-$TOMCAT_VERSION/webapps/
+        sleep 15s; tail apache-tomcat-$TOMCAT_VERSION/logs/catalina.out || true
+        curl -sS 'http://localhost:8080/vnu/?out=gnu&doc=data:text/html;charset=utf-8,%3C%21doctype%20html%3E%3Cmeta%20charset%3Dutf-8%3E%3Ctitle%3ETest%3C%2Ftitle%3E' > RESULTS
+        test -z "$(if [ -t RESULTS ]; then cat RESULTS | grep -v '^$'; fi)"
+        if [ $? != 0 ]; then cat RESULTS; fi
+        python ./checker.py check
+        python ./checker.py test
--- a/.github/workflows/darwin.yml
+++ b/.github/workflows/darwin.yml
@@ -2,8 +2,7 @@ name: macOS build for vnu
 
 on:
   push:
-    branches-ignore:
-      - "dependabot/**"
+    branches: [main]
   pull_request:
 
 jobs:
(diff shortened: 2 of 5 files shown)


.github/actions/build/action.yml
```diff
@@ -0,0 +1,28 @@
+name: 'Build'
+runs:
+  using: "composite"
```
nschonni, line 3: Interesting, I've never tried this before, but it makes sense

.github/workflows/darwin.yml
```diff
@@ -2,59 +2,48 @@ name: macOS build for vnu
...
+          python-version: '3.x'
+      - name: Update submodules
+        run: python ./checker.py update
+      - name: Install dependencies
+        run: python ./checker.py dldeps
+      - name: Build
+        uses: ./.github/actions/build
+      - name: Import GPG signing key
+        uses: crazy-max/ghaction-import-gpg@v3
+        if: github.ref == 'refs/heads/main'
```
nschonni, line 31: I'm guessing this is to avoid doing this on PRs?

.github/workflows/darwin.yml
```diff
@@ -2,59 +2,48 @@ name: macOS build for vnu
...
+        run: |
+          for file in build/dist/{*.zip,*.zip.md5,*.zip.sha1}; do gpg --yes -ab --sign $file; done || true
+          for file in build/dist/*.jar*; do gpg --yes -ab --sign $file; done || true
+          for file in build/dist-war/*.war*; do gpg --yes -ab --sign $file; done || true
+      - name: Release new macOS binary image
+        uses: "marvinpinto/action-automatic-releases@latest"
+        if: github.ref == 'refs/heads/main'
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "osx"
```
nschonni, line 48: I wonder if the old OS specific tags can be folded down into a single `develop` or something

nschonni (approved):
All looks better than the first pass.
My thought might be to create a separate "release" job that just runs on the main branch trigger. Then try and combine all the artifacts into the single tag, maybe by sequencing the build-only bits to create on tag/release.
Other future things:
- NPM release on tag
- Docker image

nschonni:
LGTM, can't see what the Docker failure is, but I'm guessing maybe it was previously broken

.github/workflows/darwin.yml
```diff
@@ -2,59 +2,54 @@ name: macOS build for vnu
...
+        if: github.ref == 'refs/heads/main'
+        with:
+          gpg-private-key: ${{ secrets.GPG_SIGNING_KEY }}
+          passphrase: ${{ secrets.GPG_SIGNING_KEY_PASSPHRASE }}
+          git-user-signingkey: true
+          git-commit-gpgsign: true
+      - name: Sign release artifacts
+        if: github.ref == 'refs/heads/main'
+        run: |
+          for file in build/dist/{*.zip,*.zip.md5,*.zip.sha1}; do gpg --yes -ab --sign $file; done || true
```
nschonni, line 40: Do the Zips need to be renamed to include the platform like `vnu.osx.zip` here before the signing?

sideshowbarker: > Do the Zips need to be renamed to include the platform like `vnu.osx.zip` here before the signing?

No, I think they’ve already got the right names. Take a look at https://github.com/sideshowbarker/validator/releases (where I have been testing this)

sideshowbarker:
> LGTM, can't see what the Docker failure is, but I'm guessing maybe it was previously broken

Yeah I can’t seem to figure out how to get rid of it. But soon I’m going to set up the Docker publishing through GitHub’s own container registry — and after I do, I’m going to remove it from Docker Hub, and delete my account at Docker Hub. Hopefully that will cause the CI integration to go away.

.github/workflows/darwin.yml
nschonni: Yeah, looks like it has the platform specific parts https://github.com/sideshowbarker/validator/releases/tag/latest
Side comment: Does it make sense to sign the other hash files (MD5/SHA1)

.github/workflows/darwin.yml
sideshowbarker:
```yaml
         if: github.ref == 'refs/heads/main' 
```
> I'm guessing this is to avoid doing this on PRs?

Yes. If there’s a better way to do that, please let me know

.github/actions/build/action.yml
sideshowbarker: > ```yaml
>  name: 'Build'
>  runs:
>    using: "composite" 
> ```
>
> Interesting, I've never tried this before, but it makes sense

I thought it was the only possible way. Is there some other way it can be done that’s more typical?

.github/workflows/darwin.yml
```diff
@@ -2,59 +2,48 @@ name: macOS build for vnu
...
+          for file in build/dist/*.jar*; do gpg --yes -ab --sign $file; done || true
+          for file in build/dist-war/*.war*; do gpg --yes -ab --sign $file; done || true
+      - name: Release new macOS binary image
+        uses: "marvinpinto/action-automatic-releases@latest"
```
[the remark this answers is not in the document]
sideshowbarker:
```yaml
         uses: "marvinpinto/action-automatic-releases@latest" 
```
> Not sure if it's better to use `1.2.1` instead of `latest`. They don't seem to be pointing at the same commit though

I just used `latest` because that’s what the examples in marvinpinto/action-automatic-releases use

.github/workflows/darwin.yml
sideshowbarker: > Take a look at [sideshowbarker/validator/releases](https://github.com/sideshowbarker/validator/releases)

By the way, the reason I didn’t turn of the separate `jar` and `linux` tag releases yet is that there’s code in some other repos depending on those. So, after we merge this, I’m going to open PRs for updating the code in those other repos to use the `latest`-tag releases instead. And once that’s done, I’ll update the repo actions and releases settings to remove those two tags/releasee

.github/actions/build/action.yml
nschonni: No, I think in the past I just did more copy/paste. This seems like a better way

.github/workflows/darwin.yml
nschonni: Probably just through splitting out the PR vs. Push jobs. I'll see what that might look like later

PR #1177, opened 2021-07-07
Drop the `jar` and `linux` release tags

sideshowbarker:
All autopublished-on-push release artifacts are now released under the tag `latest`; no more `jar`, `war`, `windows`, `linux`, `osx`


nschonni (approved):
Make sense. I can update https://github.com/validator/validator/pull/1178 after this lands

PR #1178, opened 2021-07-07
chore: Pull 'latest' tag to separate Action

nschonni:
Tried pulling the "latest" tag release to a single job, while re-using


.github/workflows/latest.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          for file in build/dist/{*.zip,*.zip.md5,*.zip.sha1}; do gpg --yes -ab --sign $file; done || true
+          for file in build/dist/*.jar*; do gpg --yes -ab --sign $file; done || true
+          for file in build/dist-war/*.war*; do gpg --yes -ab --sign $file; done || true
+      - name: Release new linux binary image
+        if: matrix.os == 'ubuntu-latest'
+        uses: "marvinpinto/action-automatic-releases@latest"
+        with:
+          repo_token: "${{ secrets.GITHUB_TOKEN }}"
+          automatic_release_tag: "linux"
```
nschonni, line 55: I guess you're dropping these in the other PR

sideshowbarker: > I guess you're dropping these in the other PR

Yeah

nschonni: Rebase and removed this

nschonni:
Hmm, looks like the commit for the tag isn't moving https://github.com/validator/validator/releases/tag/latest. Still points to the branch rename one

sideshowbarker:
> Hmm, looks like the commit for the tag isn't moving [`latest` (release)](https://github.com/validator/validator/releases/tag/latest). Still points to the branch rename one

Yeah, that’s because I had forgotten to actually add the https://github.com/validator/validator/blob/main/.github/workflows/release-prep.yml workflow to the repo. I’ve added it now.

nschonni:
OK, I tried to merge that extra job into as a dependency before the matrix

sideshowbarker:
Inspired by your approach here, I opened https://github.com/validator/validator/pull/1181 to take this further and reduce everything down to a single file. It’s not elegant, but it works.

nschonni:
@sideshowbarker I took some of your simplification work and added https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run so that the deploy stays separate, and only triggered after a successful CI run on the `main` push

sideshowbarker:
What I balk on about the approach taken in this branch and that this latest refactoring still takes is, it causes the build to be re-run 3 more times on every push. On the face of it, that seems undesirable. It roughly doubles the total build time.

Admittedly, the build is relatively quick — so in practice, I guess it just adds another 4 or 5 minutes. But it’s not clear to me what the cost of those extra 4 or 5 minutes would actually be buying us. What does it give us that the #1181 approach doesn’t have?

I guess doing it this way looks cleaner — less messy than how the #1181 code looks. And I guess it provides for a cleaner separation of the actions. But it does that at the cost of having the build code spread out over three different files, and having more repetition than #1181.

Separating the logic out across multiple files would seem desirable if it eliminates or reduces repetition, relative to the alternative (the #1181 approach). But to me at least, if we can have the logic in one single file, and at the same time minimize the repetition, that seems desirable. And that’s what I was aiming for with #1181.

sideshowbarker:
I tested this on my fork, and the results are here:

* https://github.com/sideshowbarker/validator/actions
* https://github.com/sideshowbarker/validator/releases

I needed to remove the Windows path separators to get it work on the non-Windows builds. See https://github.com/sideshowbarker/validator/commit/91952212b4b3a0a32886304f97508e2e13bff824. Otherwise it failed as in https://github.com/sideshowbarker/validator/runs/3032726474?check_suite_focus=true#step:7:11.

Also, if you look at https://github.com/sideshowbarker/validator/releases/tag/latest you’ll notice it’s not output the expected commit message there. Compare it to https://github.com/validator/validator/releases/tag/latest.

Specifically, it seems that at https://github.com/sideshowbarker/validator/blob/91952212b4b3a0a32886304f97508e2e13bff824/.github/workflows/latest.yml#L74, `${{ github.event.head_commit.message }}` is not actually giving the expected message. But I guess that’s just due to there not being an actual `github.event` context there, due to being invoked by the previous action rather than directly by the push/pull-request event.

nschonni:
I was thinking the slightly duplicated way here would be "better", so 2 more actions can be added afterwards, although I don't think it would be blocked with the other approach.
1. `npm.yaml`: "on non-latest tag" or "on release" trigger, to automagically publish to NPM for a real cut release
2. `docker.yaml`: similar to the NPM one, but for the Docker Hub (or other hub), but not on every push

nschonni:
> Specifically, it seems that at https://github.com/sideshowbarker/validator/blob/91952212b4b3a0a32886304f97508e2e13bff824/.github/workflows/latest.yml#L74, `${{ github.event.head_commit.message }}` is not actually giving the expected message. But I guess that’s just due to there not being an actual `github.event` context there, due to being invoked by the previous action rather than directly by the push/pull-request event.

Right, looks like there needs to be a step to dump `git log --pretty='format:%Creset%s' --no-merges -1` to be used for that

sideshowbarker:
> `npm.yaml`: "on non-latest tag" or "on release" trigger, to automagically publish to NPM for a real cut release

OK, I see. But so far for real cut releases, I’ve just been pushing to NPM from my local build environment. (The GitHub release creation for real cut releases also runs from my local build environment.) I’m not sure what advantage we’d get from automating the NPM publishing for real cut releases in GitHub Actions.

> `docker.yaml`: similar to the NPM one, but for the Docker Hub (or other hub), but not on every push

I have in mind to just add that to the monofile. It’s not clear to me what the win would be for having it in a separate file.

Also, by the way, I deleted the Docker Hub repo, and deleted my Docker Hub account — the immediate reason being that their ToS and policies have changed in such a way that I literally don’t understand them. And getting stuff published to Docker Hub in an automated way has always been relatively complicated and breakage-prone (witness the recent Docker-related CI failures we were seeing). So I couldn’t seen any value in continuing to try to wrestle with their complications.

So what I’ve done instead is, I moved the Docker image publishing completely over to https://github.com/validator/validator/pkgs/container/validator (aka https://ghcr.io/validator/validator), and updated the docs. See https://github.com/validator/validator/commit/ef729417834ccb3010cfd19a89502a61b871847d.

nschonni:
2 other thoughts (probably doable by both)
- Create a separate action to publish the website
- What should be ignored from triggering CI/release https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths. Can be done in the mono-job, but would have to match for all cases

PR #1181, opened 2021-07-09
CI: Switch to putting all build steps in a single build.yml file

sideshowbarker:


.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Log in to the container registry
+        uses: docker/login-action@f054a8b539a109f9f41c372932f1ae047eff08c9
```
nschonni, line 20:
```suggestion
        uses: docker/login-action@v1.10.0
```

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
```
nschonni, line 28:
```suggestion
        uses: docker/metadata-action@v3.4.0
```

.github/workflows/docker-push.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          images: ghcr.io/validator/validator
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
```
nschonni, line 33:
```suggestion
        uses: docker/build-push-action@v2.6.1
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Install dependencies
+        run: python checker.py dldeps
+      - name: Build
+        run: |
+          python checker.py build
+          python checker.py jar
+          python checker.py image
+          python checker.py war
+      - name: Test
+        shell: bash
```
nschonni, line 65: I think this wasn't running for Windows before, but I guess it's OK with the explicit `bash` environment?

.github/workflows/build.yml
```diff
@@ -0,0 +1,121 @@
...
+           git config user.email github-actions@github.com
+           git push origin :refs/tags/latest || true
+           git tag -m latest -fa latest || true
+           git push origin main --tags
+  build:
+    name: Build on ${{ matrix.os }} with Java ${{ matrix.java }}
+    needs: prep
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
```
nschonni, line 27: This might be one were the separate action is slightly better, because you probably want to see all the failures during a PR's CI, but fast-fail when pushing the built tags. Could probably get something similar with a condition, or just leave it as-is
```suggestion
      fail-fast: false
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,121 @@
...
+        if: github.ref == 'refs/heads/main'
+        run: |
+           git config user.name github-actions
+           git config user.email github-actions@github.com
+           git push origin :refs/tags/latest || true
+           git tag -m latest -fa latest || true
+           git push origin main --tags
+  build:
+    name: Build on ${{ matrix.os }} with Java ${{ matrix.java }}
+    needs: prep
```
nschonni, line 24: I think this is probably OK, because you have the conditional on the step in the other job, but if you made the whole job conditional, it would likely stall

nschonni (approved):

.github/workflows/build.yml
sideshowbarker: > I think this is probably OK, because you have the conditional on the step in the other job, but if you made the whole job conditional, it would likely stall

OK

.github/workflows/build.yml
sideshowbarker: > I think this wasn't running for Windows before, but I guess it's OK with the explicit `bash` environment?

Right, yes. And I guess running it in bash on Windows is kind of not actually modeling/testing what a normal user in their normal environment on Windows would be doing — but anyway that’s only the case for this one test step, which itself is anyway testing the runtime behavior and not just the build success. So it seems OK to use bash on Windows for this one part.

PR #1235, opened 2021-09-13
Add CodeQL Action

XhmikosR:
Currently analyzing JavaScript and Python; Java should be added later too


sideshowbarker (approved):

XhmikosR:
@sideshowbarker might be worth looking into hooking up Java too later. I'm not familiar with it, but it shouldn't be too hard to add.

Issue #1241, opened 2021-09-23
Update allowances for wbr and br elements


scottaohara:
The *`wbr` element* presently allows all roles and global aria-* attributes.  This was an oversight that will be corrected in the next version of the ARIA in HTML specification.

The allowed roles should now only be `none` and `presentation`.
The only allowed aria-* attribute should be `aria-hidden`.

Additionally, the *`br`* element also needs to be updated to remove the allowance for all global aria attributes – instead only allowing for the `aria-hidden` attribute. This will then match the allowances for the `wbr` element, described above.

PR #1251, opened 2021-10-08
update the allowed roles for wbr

scottaohara:
update allowed roles for `wbr` to match `br` element

related to #1241, but does not (presently) address updating `wbr` and `br` to only allow `aria-hidden`


stevefaulkner:
@sideshowbarker  is there a reason this is still open?

scottaohara:
The PR only addressed part of the original issue, concerning the allowed roles.  the only allowed attribute should be aria-hidden.  i don't think that's been addressed yet?

sideshowbarker:
Yeah, since this PR wasn’t marked as fixing or closing #1241, then that issue remains open

PR #24, opened 2015-01-10
Added gulp plugin link

watilde:
Hi

I love to use this validator. So I created a gulp plugin using vnu.jar, could I add a link on this page? I'd be happy to do that.

You can see the preview at this url: http://watilde.github.io/validator/index.html

Thanks.

PR #1400, opened 2022-07-12
Add multiplatform docker support

Jubeki:
Currently the docker image is only build for amd64, with this change it will also build for arm64.

Steps are taken from the official repository of the already used Docker Build Push Action:
https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

Note:
I am not an expert for Docker images or Multiplatform support, and only tried to apply worthwhile changes.

### DIFF
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -45,5 +45,6 @@ jobs:
         with:
           context: .
           push: true
+          platforms: linux/amd64,linux/arm64
           tags: ghcr.io/validator/validator:latest, ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -17,19 +17,29 @@ jobs:
       packages: write
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
       - name: Log in to the container registry
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Extract metadata (tags, labels)
         id: meta
         uses: docker/metadata-action@v4
         with:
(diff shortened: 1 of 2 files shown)


sideshowbarker:
Thanks much. I’ll go ahead and merge this. If it doesn’t work as expected, we can always back it out and try again.

sideshowbarker:
OK, yeah, this doesn’t seem to have worked as expected:

https://github.com/validator/validator/runs/7332356544?check_suite_focus=true#step:7:750

> #24 ERROR: failed to calculate checksum of ref s11b53u52xa4ql6dicaf8r2se::raw4hyveuqz4nacd0eze0awr4: "/lib/x86_64-linux-gnu/libz.so.1": not found

Jubeki:
@sideshowbarker 
Yesterday I tried my best to find a fix for that. I have an idea, but I am not completely sure how to implement that. 

The Problem lies with the following line in the Dockerfile: https://github.com/validator/validator/blob/9e4252fd390319a50230a92f2b4b4a741a1bbec7/Dockerfile#L26

This can simply be fixed by copying another directory (only needed for the arm image):
```diff
- COPY --from=builder /lib/x86_64-linux-gnu/libz.so.1 /lib/x86_64-linux-gnu/libz.so.1
+ COPY --from=builder /lib/aarch64-linux-gnu/libz.so.1 /lib/aarch64-linux-gnu/libz.so.1 
```

This would be possible with creating another Dockerfile for the arm version and then pushing both amd64 and arm64 to the ghcr.io registry. After that you would create a docker-manifest which includes both the amd64 and arm64 version and push that to the latest tag.

But there is still a Problem with the arm version:
https://github.com/validator/validator/blob/9e4252fd390319a50230a92f2b4b4a741a1bbec7/Dockerfile#L36

vnu-runtime-image is a precompiled amd64 version, which means, this would still need to be emulated by docker, which would result in a amd64 docker image again.

The only idea to fix this, would be either to compile the vnu-runtime-image also in an arm version, or to directly execute the vnu.jar and starting the Webserver there, without using a precompiled version.

This may be achievable using `gcr.io/distroless/java11`, but there are some other implications, like that it can only execute a jar file with the `-jar` option, but not using `-cp`.
Another option would be creating an image using Tomcat and deploying the vnu.war but I don't have much knowledge about that.

sideshowbarker:
I’ve reverted this for now.

I’d be happy to try out an alternative patch with a fix (e.g., as described in https://github.com/validator/validator/pull/1400#issuecomment-1185310331).

PR #1444, opened 2022-10-19
imagesrcset and imagesizes in link

### LINKED ISSUE #826: Allow imagesrcset and imagesizes attributes for link rel=preload
Currently only supported in Blink.

* https://www.chromestatus.com/feature/5164259990306816
* https://github.com/whatwg/html/commit/d11bc53d8c82b8c3f761501f9bc98f2a13a5f239
* https://html.spec.whatwg.org/multipage/semantics.html#attr-link-imagesrcset
* https://html.spec.whatwg.org/multipage/semantics.html#attr-link-imagesizes

### LINKED ISSUE #1255: Validator reports LINK with 'imagesrcset' but no 'href' as invalid, although HTML docs indicate it would be valid
The validator throws this error when a LINK element with an imagesrcset attribute does not also have a href:

```
Error: Element link is missing one or more of the following attributes: href, resource.

From line 1, column 4097; to line 1, column 4600

<link rel=preload as=image imagesrcset="images/home/landing-image-set/a02/a02-683px.20210331021121.j…ng-image-set/a02/a02-1225px.20210331021121.jpg 1225w" imagesizes="(max-width: 579.98px) 85vw, 50vw">

```

The HTML specs on the rel=preload imagesrcset LINK element found here:

https://html.spec.whatwg.org/multipage/semantics.html#attr-link-imagesrcset

...includes this example:

```
 These attributes allow preloading the appropriate resource that is later used by an img element that has the corresponding values for its srcset and sizes attributes:
 
 <link rel="preload" as="image"
       imagesrcset="wolf_400px.jpg 400w, wolf_800px.jpg 800w, wolf_1600px.jpg 1600w"
       imagesizes="50vw">
 
 <!-- ... later, or perhaps inserted dynamically ... -->
 <img src="wolf.jpg" alt="A rad wolf"
      srcset="wolf_400px.jpg 400w, wolf_800px.jpg 800w, wolf_1600px.jpg 1600w"
      sizes="50vw">
 
 Note how we omit the href attribute, as it would only be relevant for browsers that do not support imagesrcset, and in those cases it would likely cause the incorrect image to be preloaded.

```

Also see: https://html.spec.whatwg.org/multipage/semantics.html#the-link-element

...where it states:

> One or both of the href or imagesrcset attributes must be present.

The validator should accept LINKs with either href or imagesrcset.

NiedziolkaMichal:
This PR adds imagesrcset & imagesizes attributes to `<link>`. Validation error will be printed when imagesrcset or imagesizes attributes are present, while rel is not preload or as is not image. Attribute imagesizes can only be present, while imagesrcset uses width descriptor.
Attribute href is no longer required, as long as imagesrcset is present. When I have moved this check to Assertions file, rdfa tests started to fail. They contain following links: `<link property="rdfa:copy" resource="_:a">`. I don't really understand rdfa, so in Assertions I have allowed to skip href or imagesrcset attributes, as long as resource is present.
This PR also adds 4 tests, for which there is a separate PR: [#4](https://github.com/validator/tests/pull/4)

Fixes #826
Fixes #1255

### DIFF
--- a/schema/html5/meta.rnc
+++ b/schema/html5/meta.rnc
@@ -142,8 +142,8 @@ datatypes w = "http://whattf.org/datatype-draft"
 		&	common.attrs.i18n
 		&	common.attrs.present
 		&	common.attrs.other
-		&	link.attrs.href
 		&	link.attrs.rel
+		&	link.attrs.href?
 		&	link.attrs.as?
 		&	link.attrs.integrity?
 		&	shared-hyperlink.attrs.hreflang?
--- a/schema/html5/meta.rnc
+++ b/schema/html5/meta.rnc
@@ -156,6 +156,8 @@ datatypes w = "http://whattf.org/datatype-draft"
 		&	link.attrs.scope?
 		&	link.attrs.updateviacache?
 		&	link.attrs.workertype?
+		&	link.attrs.imagesrcset?
+		&	link.attrs.imagesizes?
 		#	link.attrs.title included in common.attrs
 		&	embedded.content.attrs.crossorigin?
 		&	common.attrs.aria.role.link?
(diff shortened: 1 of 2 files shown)


sideshowbarker (approved):

XhmikosR:
@sideshowbarker is this fixed in the latest version which ships with npm? I just tried it and it seems to fail with vnu-jar 22.9.29 https://github.com/twbs/blog/actions/runs/4427134999/jobs/7764312916

PR #1455, opened 2022-10-26
Corrected possible attributes for input elements

NiedziolkaMichal:
This PR slightly reworks web-forms.rnc & web-forms2.rnc, correcting list of possible form attributes for each input type. It made usage of INPUT_ATTRIBUTES in Assertions.java obsolete and was removed. Now any invalid attribute in input element shows error like that:
image

I have also cleaned up both rnc files a little bit, moving all attributes to web-forms.rnc & connecting input types which were present in both files.

I think it would be better to make web-forms-input.rnc and move all rules regarding input types in there. In web-forms2.rnc there would be just few rules, so we could be move them to web-forms.rnc and delete the file.
We could also change naming of types regarding form attributes, because currently they are under:
1. shared-form.attrs
2. common-form.attrs
3. input.attrs

What do you think?

### DIFF
--- a/schema/html5/web-forms.rnc
+++ b/schema/html5/web-forms.rnc
@@ -4,12 +4,55 @@ datatypes w = "http://whattf.org/datatype-draft"
 ##  RELAX NG Schema for HTML 5: Web Forms 1.0 markup                  #
 # #####################################################################
 
-## Shared attributes for form controls
+## Groups of shared attributes
 
+# Shared by all form controls
 	common-form.attrs =
 		(	common-form.attrs.name?
 		&	common-form.attrs.disabled?
+		&	common-form.attrs.form?
 		)
+
+# Shared by all single or multiple line text edit controls
+	shared-form.attrs.text-edit =
+		(	common-form.attrs
+		&	shared-form.attrs.minlength?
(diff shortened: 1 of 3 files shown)


sideshowbarker:
I’m going ahead and closing this unmerged — somewhat with regrets, because I think objectively speaking it’s a decent change. But subjectively speaking, in my judgement, the current form of the error message:

> Attribute list is only allowed when the input type is color, date, datetime-local, email, month, number, range, search, tel, text, time, url, or week.

…is better for users — across the set of environments where they may see the messages — than the new message this change would cause to be emitted instead.

I recognize that with this change, the UI of the results page in the checker UI will show the detailed _“Attributes for element input”_ info — but users of the CLI who run vnu in a shell environment will not see that info.

It therefore seems better for me to keep the details within the actual error message itself.

The very over-complicated nature of the `input` element unfortunately leads to a lot of complications in how the checker handles error reporting for the `input` element — and can make it difficult to decide how best to frame the requirements for users.

So I’m not trying to assert I’ve got things absolutely right in deciding not to merge this PR. It’s just a judgement call, and my judgement in this case leans toward concluding that keeping things the way they are now for this particular case is on balance the better thing.

PR #1466, opened 2022-10-28
blocking="render"

### LINKED ISSUE #1315: Implement `blocking=render` validation
URL of WHATWG spec or PR:
https://github.com/whatwg/html/pull/7474

**Describe the solution you'd like**
No error  for documents using `blocking=render` on `link`, `script`, `style`

Error for documents using the `blocking` attribute incorrectly, as per spec.

NiedziolkaMichal:
Adds implementation for [blocking](https://github.com/whatwg/html/pull/7474/files) attribute to `link`, `style` & `script`. When used inside `link` element, it requires `rel` attribute to be either `modulepreload`, `preload`, or `stylesheet`.

Fixes #1315

### DIFF
--- a/schema/html5/common.rnc
+++ b/schema/html5/common.rnc
@@ -212,7 +212,7 @@ common.attrs.other =
 
 	common.attrs.blocking = 
 		attribute blocking {
-			w:string "render" | w:string ""
+			list { w:string "render" }
 		}
 
 	common.attrs.nonce =
--- a/schema/html5/common.rnc
+++ b/schema/html5/common.rnc
@@ -210,6 +210,11 @@ common.attrs.other =
 			w:string "autofocus" | w:string ""
 		}
 
+	common.attrs.blocking = 
+		attribute blocking {
+			w:string "render" | w:string ""
+		}
+
 	common.attrs.nonce =
 		attribute nonce {
 			string
--- a/src/nu/validator/checker/schematronequiv/Assertions.java
+++ b/src/nu/validator/checker/schematronequiv/Assertions.java
@@ -3166,6 +3166,17 @@ public class Assertions extends Checker {
                             + " \u201Cprerender\u201D, or"
                             + " \u201Cstylesheet\u201D.");
                 }
+                if (atts.getIndex("", "blocking") > -1
+                        && atts.getValue("", "blocking").equalsIgnoreCase("render")
+                        && (atts.getIndex("", "rel") == -1
+                        || !(relList.contains("modulepreload")
+                            || relList.contains("preload")
+                            || relList.contains("stylesheet")))) {
+                    err("A \u201Clink\u201D element with a"
+                                + " \u201Cblocking\u201D attribute containing value \u201Crender\u201D,"
+                                + " must have a \u201Crel\u201D attribute with value \u201Cmodulepreload\u201D," +
+                                " \u201Cpreload\u201D or \u201Cstylesheet\u201D.");
+                }
             }
 
             // microdata
--- a/src/nu/validator/checker/schematronequiv/Assertions.java
+++ b/src/nu/validator/checker/schematronequiv/Assertions.java
@@ -3167,13 +3167,12 @@ public class Assertions extends Checker {
                             + " \u201Cstylesheet\u201D.");
                 }
                 if (atts.getIndex("", "blocking") > -1
-                        && atts.getValue("", "blocking").equalsIgnoreCase("render")
                         && (atts.getIndex("", "rel") == -1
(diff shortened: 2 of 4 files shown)


schema/html5/common.rnc
```diff
@@ -210,6 +210,11 @@ common.attrs.other =
 			w:string "autofocus" | w:string ""
 		}
 
+	common.attrs.blocking = 
+		attribute blocking {
+			w:string "render" | w:string ""
```
sideshowbarker, line 215:
```suggestion
			w:string "render"
```
Per https://html.spec.whatwg.org/multipage/urls-and-fetching.html#blocking-attribute the empty string is not actually allowed as a `blocking` value, right?

src/nu/validator/checker/schematronequiv/Assertions.java
```diff
@@ -3157,6 +3157,17 @@ else if ("bdo" == localName && atts.getIndex("", "dir") < 0) {
                             + " \u201Cprerender\u201D, or"
                             + " \u201Cstylesheet\u201D.");
                 }
+                if (atts.getIndex("", "blocking") > -1
+                        && atts.getValue("", "blocking").equalsIgnoreCase("render")
```
sideshowbarker, line 3161: Per https://html.spec.whatwg.org/multipage/semantics.html#attr-link-blocking the spec states:
> The blocking attribute is a [blocking attribute](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#blocking-attribute). It is used by link types [modulepreload](https://html.spec.whatwg.org/multipage/links.html#link-type-modulepreload), [preload](https://html.spec.whatwg.org/multipage/links.html#link-type-preload) and [stylesheet](https://html.spec.whatwg.org/multipage/links.html#link-type-stylesheet), and it must only be specified on link elements that have a [rel](https://html.spec.whatwg.org/multipage/semantics.html#attr-link-rel) attribute containing one of those keywords.

That means we don’t need to check the value of the `blocking` attribute here, right? — Because the spec states the attribute itself (regardless of its value) must only be used with elements that also have one of the given `rel` values.

sideshowbarker (changes requested):

schema/html5/common.rnc
NiedziolkaMichal: I think it is. [Blocking attribute definition](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#blocking-attributes) says "A [blocking attribute](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#blocking-attribute) must have a value that is an [unordered set of unique space-separated tokens](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#unordered-set-of-unique-space-separated-tokens)", while definition of set is "A set of space-separated tokens is a string containing zero or more words (known as tokens) separated by one or more [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace), where words consist of any string of one or more characters, none of which are [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).". So zero words seems to be a valid value of blocking attribute. Do you agree?

src/nu/validator/checker/schematronequiv/Assertions.java
NiedziolkaMichal: You are right. I removed this part of the condition and edited message.

schema/html5/common.rnc
sideshowbarker:
```suggestion
			list { w:string "render" }
```
This discussion made me realize that the checker schema shouldn’t be restricting the value to string, but instead needs to allow a list — since the spec says it can be a set of space-separated tokens, not just a single string.

And because there’s no way for a set of space-separated tokens to include the empty string, it wouldn’t make any sense anyway for the checker schema to include the empty string in the set of tokens allowed in the list.

And on top of that, the spec doesn’t just say it’s “unordered set of unique space-separated tokens” — instead it says it’s an “unordered set of unique space-separated tokens _**each of which are [possible blocking tokens](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#possible-blocking-token)**_”

And the definition above that says:

> The operations that can be blocked are represented by **possible blocking tokens**, which are strings listed by the following table:

…and that table lists only one single allowed “possible blocking token”, which is `"render" `. It explicitly does not list the empty string as a “possible blocking token” — and the spec wouldn’t make any sense if it did allow the empty string as a token, because there’s no way for a set of space-separated tokens to include the empty string.

NiedziolkaMichal: That sounds good. I have commited your suggested change.

sideshowbarker (approved):

Issue #1106, opened 2021-03-06
disabled & aria-disabled validation test cases


scottaohara:
per these [test cases](https://w3c.github.io/html-aria/tests/disabled-test.html), would expect the validator to report the following:

test 1: results are largely as expected - validator throws no warnings/errors.  Exception should be that `input type=hidden aria-disabled=true"` should throw an error as no aria-* attributes are allowed on this input type.  (however, as this element is not exposed to users, I can understand if this is a "wontfix" issue.)

test 2: flags warnings as expected and needs no updates.

test 3: *nothing presently reported by validator* - should report errors for use of `aria-disabled=false` w/ `disabled` attribute, as they are in opposition and `disabled` attribute wins.

sideshowbarker:
> per these [test cases](https://w3c.github.io/html-aria/tests/disabled-test.html), would expect the validator to report the following:
> 
> test 1: results are largely as expected - validator throws no warnings/errors. Exception should be that `input type=hidden aria-disabled=true"` should throw an error as no aria-* attributes are allowed on this input type. (however, as this element is not exposed to users, I can understand if this is a "wontfix" issue.)

Fixed by https://github.com/validator/validator/commit/84bde80

> test 3: _nothing presently reported by validator_ - should report errors for use of `aria-disabled=false` w/ `disabled` attribute, as they are in opposition and `disabled` attribute wins.

Fixed by https://github.com/validator/validator/commit/726e759

PR #1468, opened 2022-10-29
Fix check for input type=hidden using any aria

NiedziolkaMichal:
This is correction for commit [84bde80](https://github.com/validator/validator/commit/84bde80) fixing #1106. Now assertion checks type=hidden instead of global attribute hidden. I have also removed 2 lines regarding aria-hidden, because `<input type=hidden aria-hidden="true"` would print:

image


sideshowbarker (approved):
oops — nice catch

PR #1471, opened 2022-10-31
Headings in nested sections

### LINKED ISSUE #1272: Incorrect warning about missing heading when an Article has Aside as the first child
For the following content:

```
<!DOCTYPE html><html lang="en-GB"><head><title>Test</title></head><body>

<article>
 <aside>This article was nominated best article of 2020s</aside>
 <h1>Identifying header</h1>
 <p>Content</p>
</article>

</body></html>
```

There is a warning that the article lacks an identifying header. 

- The issue is not having an aside (if the aside is moved to come after the h1, the warning disappears).
- The issue is also not that there is content before the h1 (if there is a paragraph element before the h1, there is no warning).

So there appears to be a warning specifically when an aside element appears before the header. Nothing in the specifications of either aside or article elements seems to justify this.

NiedziolkaMichal:
Fixes issue of heading elements not always being found while sectioning element was nested.

Fixes #1272

### DIFF
--- a/src/nu/validator/checker/schematronequiv/Assertions.java
+++ b/src/nu/validator/checker/schematronequiv/Assertions.java
@@ -24,7 +24,17 @@ package nu.validator.checker.schematronequiv;
 
 import java.io.ByteArrayInputStream;
 import java.io.StringReader;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


src/nu/validator/checker/schematronequiv/Assertions.java
```diff
@@ -24,17 +24,8 @@
 
 import java.io.ByteArrayInputStream;
 import java.io.StringReader;
+import java.util.*;
```
sideshowbarker, line 27:
```suggestion
import java.util.*;
```
We intentionally have never used any wildcard imports in the sources for the checker (and the htmlparser); it’s not an accident or oversight that we haven’t.

So unless there’s some very compelling reason to use a wildcard import here, let’s please instead stick with explicitly listing everything we need.

NiedziolkaMichal: That's just automatic change by my IDE. I have commited a change.

sideshowbarker (approved):
Nice work

PR #1475, opened 2022-11-02
Error on nested "a" in SVG

### LINKED ISSUE #1081: nested <a> in <svg> pass
Code to reproduce:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width">
  <title>nested links in svg</title>
</head>
<body>
<svg width="5cm" height="3cm" viewBox="0 0 5 3" version="1.1"
     xmlns="http://www.w3.org/2000/svg">
  <desc>Example link01 - a link on an ellipse
  </desc>
  <a href="http://www.svgwg.org">
    <rect x=".01" y=".01" width="4.98" height="2.98" 
          fill="none" stroke="blue"  stroke-width=".03"/>
    <a href="http://www.w3.org">
      <ellipse cx="2.5" cy="1.5" rx="2" ry="1"
               fill="red" />
    </a>
  </a>
</svg>
</body>
</html>
```

This sample passes at https://validator.w3.org/nu/#textarea

If I understand the spec situation correctly, it should have an error for nested anchors. (Or at least https://html.spec.whatwg.org/#svg-0 links to https://svgwg.org/svg2-draft/ which says [nested anchors are no allowed](https://svgwg.org/svg2-draft/linking.html#Links).)

NiedziolkaMichal:
Prints error when "a" element inside SVG has "a" descendant in that SVG or anywhere outside of it. Setting "name" variable in StackNode class of non html elements is potentially a breaking change, but I think everything will be fine.

Fixes #1081

### DIFF
--- a/src/nu/validator/checker/schematronequiv/Assertions.java
+++ b/src/nu/validator/checker/schematronequiv/Assertions.java
@@ -1734,6 +1734,16 @@ public class Assertions extends Checker {
             warn("Element \u201Cmath\u201D does not need a"
                     + " \u201Crole\u201D attribute.");
         }
+        if ("http://www.w3.org/2000/svg" == uri) {
+            if ("a" == localName) {
+                for(int i = 0;i < currentPtr;i++) {
+                    String ancestorName = stack[currentPtr - i].getName();
+                    if("a".equals(ancestorName)) {
+                        err("The element \u201Ca\u201D must not appear as a descendant of an element \u201Ca\u201D.");
+                    }
+                }
+            }
+        }
         if ("http://www.w3.org/1999/xhtml" == uri) {
             boolean controls = false;
             boolean hidden = false;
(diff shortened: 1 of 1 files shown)


src/nu/validator/checker/schematronequiv/Assertions.java
```diff
@@ -1734,6 +1734,16 @@ public void startElement(String uri, String localName, String name,
             warn("Element \u201Cmath\u201D does not need a"
                     + " \u201Crole\u201D attribute.");
         }
+        if ("http://www.w3.org/2000/svg" == uri) {
+            if ("a" == localName) {
```
nschonni, line 1738: Should the `if`s be combined?

NiedziolkaMichal: Alright. I have updated this condition.

nschonni: Sorry if this was too nit picky, I'm not sure if the original was in line with the prevailing style of the codebase, just stuck out for me

NiedziolkaMichal: Actually your suggested change is more in style of the codebase. I have initially separated those conditions, because I expect more assertions be added for SVG.

sideshowbarker (approved):

PR #1721, opened 2024-05-03
Support all official CSP/iframe sandbox parameters

### LINKED ISSUE #1720: False positive on valid CSP `sandbox` values
The Nu Validator has an incomplete dictionary of valid values for the `sandbox` CSP directive and `iframe` attribute.

Relevant WHATWG spec URL: https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-sandbox

`allow-top-navigation-to-custom-protocols` is missing from both; additional directives such as `allow-downloads` are missing from the CSP `sandbox` directive. I'll make a PR soon, though in the long run it may be worth enumerating these in one place instead of in two places with different syntaxes.

Seirdy:
Sync both CSP and iframe sandbox parameters to the latest spec. Add these keywords to the CSP sandbox directive:

- allow-downloads
- allow-orientation-lock
- allow-presentation
- allow-top-navigation-by-user-activation
- allow-top-navigation-to-custom-protocols

Add these keywords to the iframe sandbox attribute:

- allow-top-navigation-to-custom-protocols

Relevant specification URLs:

Valid values for sandbox attributes:
<https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-sandbox>

CSP sandbox directive deferring to WHATWG HTML Living Standard: <https://www.w3.org/TR/CSP3/#directive-sandbox>

Fixes #1720

### DIFF
--- a/src/nu/validator/datatype/ContentSecurityPolicy.java
+++ b/src/nu/validator/datatype/ContentSecurityPolicy.java
@@ -45,8 +45,11 @@ public class ContentSecurityPolicy extends AbstractDatatype {
 
     private static final String SANDBOX_KEYWORDS = "("
             + "allow-forms|allow-modals|allow-pointer-lock"
+            + "|allow-downloads|allow-orientation-lock|allow-presentation"
             + "|allow-popups-to-escape-sandbox|allow-popups|allow-same-origin"
-            + "|allow-scripts|allow-top-navigation)";
+            + "|allow-scripts|allow-top-navigation"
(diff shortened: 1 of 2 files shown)


sideshowbarker (approved):

sideshowbarker:
Thanks much — there will be a delay in getting this pushed to https://validator.w3.org/nu/, but I’ll post another comment here after it has actually been deployed there

sideshowbarker:
This change has been deployed to https://validator.w3.org/nu/

PR #1723, opened 2024-05-09
Update jetty to 11.0.20

marcospereira:
This required updating a few transitive dependencies:

- commons-fileupload to version 2.0.0-M2: which supports the Jakarta Servlet API
- commons-io to version 2.15.1

Also, code references to `javax.servlet` were updated to use `jakarta.servlet`.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,5 +18,9 @@ indent_size = 2
 [*.{js,css,xml}]
 indent_style = tab
 
+[**/build.xml]
+indent_style = space
+indent_size = 2
+
 [*.md]
 trim_trailing_whitespace = false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,12 +27,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        java: [22, 21, 17, 11.0.3, 8.0.192]
+        java: [22, 21, 17, 11.0.23, 8.0.192]
         exclude:
           - os: macos-latest
             java: 8.0.192
           - os: macos-latest
-            java: 11.0.3
+            java: 11.0.23
           - os: windows-latest
             java: 8.0.192
           - os: windows-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,14 +27,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        java: [22, 21, 17, 11.0.23, 8.0.192]
+        java: [22, 21, 17, 11.0.23]
         exclude:
           - os: macos-latest
-            java: 8.0.192
-          - os: macos-latest
             java: 11.0.23
-          - os: windows-latest
-            java: 8.0.192
           - os: windows-latest
             java: 11.0.3
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ or changes to any options/interfaces the checker exposes for developers.
 
 # NN.NN.NN
 NN XXX NNNN
+  - Require Java 11; stop supporting Java 8
   - Report void elements that have a trailing slash in their start tags
   - Skip template contents when checking for duplicate IDs (bug fix)
   - Allow `loading` attribute for the `iframe` element (lazy loading)
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ It is released upstream in these formats:
   * pre-compiled Linux, Windows, and macOS binaries that include an embedded
   Java runtime
 
-  * `vnu.jar` — a portable version you can use on any system that has Java 8 or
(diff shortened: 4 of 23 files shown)


sideshowbarker:
Thanks for the patch. I’m enthusiastic about merging this, but can’t yet because the change as currently written seems to break the build of the binary image. See the CI checks for the branch; for example, at https://github.com/validator/validator/actions/runs/9019893225/job/24795094469?pr=1723#step:7:646 —
> Error: .\build\dist\vnu.jar is a modular JAR file that cannot be specified with the --generate-module-info option

I don’t have any idea why the change in the PR makes that happen, and I don’t have any idea how to fix it.

The build for the binary images is something that I set up quite a long time ago, and I recall it required some finessing to make it work. But since then, it’s just worked and I’ve not needed to fiddle with it again.

marcospereira:
Hey @sideshowbarker, thanks for triggering the workflows. :-)

> > Error: .\build\dist\vnu.jar is a modular JAR file that cannot be specified with the --generate-module-info option
> 
> I don’t have any idea why the change in the PR makes that happen, and I don’t have any idea how to fix it.

I suspect this is happening because the new dependency jars include a `module-info.class`, and that is confusing the `jdeps` command when generating a `module-info.java`.

Also, I noticed a few entries were missing in the `war` target. There are a few new commits to move this forward, but the `image` task still fails on my machine, but with a different error now:

```
Error: Missing dependencies: classes not found from the module path and classpath.
To suppress this error, use --ignore-missing-deps to continue.

vnu
   javax.portlet.ClientDataRequest                    -> javax.servlet.http.Part                            not found
   javax.portlet.PortletRequest                       -> javax.servlet.http.Cookie                          not found
   javax.portlet.PortletResponse                      -> javax.servlet.http.Cookie                          not found
   javax.portlet.ResourceRequest                      -> javax.servlet.DispatcherType                       not found
   javax.portlet.filter.ClientDataRequestWrapper      -> javax.servlet.http.Part                            not found
   javax.portlet.filter.PortletRequestWrapper         -> javax.servlet.http.Cookie                          not found
   javax.portlet.filter.PortletResponseWrapper        -> javax.servlet.http.Cookie                          not found
   javax.portlet.filter.ResourceRequestWrapper        -> javax.servlet.DispatcherType                       not found
   org.apache.commons.logging.impl.ServletContextCleaner -> javax.servlet.ServletContextEvent                  not found
   org.apache.commons.logging.impl.ServletContextCleaner -> javax.servlet.ServletContextListener               not found
   org.apache.log.output.ServletOutputLogTarget       -> javax.servlet.ServletContext                       not found
Traceback (most recent call last):
  File "/Users/marcospereira/src/github.com/validator/validator/checker.py", line 42, in <module>
    build.main(args)
  File "/Users/marcospereira/src/github.com/validator/validator/build/build.py", line 2016, in main
    release.createRuntimeImage()
  File "/Users/marcospereira/src/github.com/validator/validator/build/build.py", line 1069, in createRuntimeImage
    f = open(os.path.join(self.vnuModuleInfoDir, "module-info.java"), 'r+')
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: './build/dist/vnu/module-info.java'
```

The jar works as expected, though. Running `java -cp build/dist/vnu.jar nu.validator.servlet.Main 8888` starts the server as expected, and I can use the application. Although I didn't test it extensively.

sideshowbarker:
> Hey @sideshowbarker, thanks for triggering the workflows. :-)
> 
> > > Error: .\build\dist\vnu.jar is a modular JAR file that cannot be specified with the --generate-module-info option
> > 
> > 
> > I don’t have any idea why the change in the PR makes that happen, and I don’t have any idea how to fix it.
> 
> I suspect this is happening because the new dependency jars include a `module-info.class`, and that is confusing the `jdeps` command when generating a `module-info.java`.

Aha, yeah — I think that would do it, yes

> Also, I noticed a few entries were missing in the `war` target. There are a few new commits to move this forward, but the `image` task still fails on my machine, but with a different error now:
> …
> Error: Missing dependencies: classes not found from the module path and classpath.
> To suppress this error, use --ignore-missing-deps to continue.

So yeah, maybe we can “fix” the problem just by adding that `--ignore-missing-deps` flag, I dunno

> The jar works as expected, though. Running `java -cp build/dist/vnu.jar nu.validator.servlet.Main 8888` starts the server as expected, and I can use the application. Although I didn't test it extensively.

But using `java -cp build/dist/vnu.jar nu.validator.servlet.Main 8888` isn’t running the image, right?

To run the image, you’d instead need to do:

```
vnu-runtime-image/bin/java      nu.validator.servlet.Main 8888
```

sideshowbarker (approved):

sideshowbarker:
> So yeah, maybe we can “fix” the problem just by adding that `--ignore-missing-deps` flag, I dunno

It turns out that we can. So CI is all green here now, so I’m going ahead and merging this. If there are any further changes needed, we can handle them in further follow-up PRs.

Thanks much

Issue #1737, opened 2024-07-30
build.py ignores exit status of subprocesses


kevinoid:
There are several locations in `build/build.py` where the return value of `runCmd()` is ignored.  This makes it difficult for callers to determine whether the requested operation succeeded (e.g. `build.py update` or `build.py docker-build`) and cause operations to silently proceed despite errors (e.g. `build.py all` runs the validator even if `update` failed).

I'd like to propose changing `runCmd()` to call [`subprocess.check_call`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) so that `runCmd()` will raise an exception for non-zero exit codes and exit with a non-zero status if the exception is not caught.  Would a pull request with this change be acceptable?  Are there any uses of `runCmd()` where non-zero exit status should be ignored?  (It does not appear so to me.)

Thanks for considering,
Kevin

sideshowbarker:
> I'd like to propose changing `runCmd()` to call [`subprocess.check_call`](https://docs.python.org/3/library/subprocess.html#subprocess.check_call) so that `runCmd()` will raise an exception for non-zero exit codes and exit with a non-zero status if the exception is not caught. Would a pull request with this change be acceptable?

Yes, sounds good to me

> Are there any uses of `runCmd()` where non-zero exit status should be ignored? (It does not appear so to me.)

Same here. I can’t think of any

PR #1739, opened 2024-08-01
build: raise `CalledProcessError` for failed subprocesses

kevinoid:
As described in #1737, `build.py` does not always check the exit code of the subprocesses it calls via `runCmd()`.  This makes it difficult for callers to determine whether the requested operation succeeded (e.g. `build.py update` or `build.py docker-build`) and cause operations to silently proceed despite errors (e.g. `build.py all` runs the validator even if update failed).

Change `runCmd()` to use `subprocess.check_call()` instead of `subprocess.call()` so that `CalledProcessError` will be raised if the subprocess exits with non-zero status.  Remove conditional exit on `runCmd()` result, since `CalledProcessError` is now raised in this case and will cause the process to exit with a non-zero code due to an unhandled exception.

Fixes: #1737

Thanks for considering,
Kevin

P.S, If you'd prefer that `build.py` produce less technical/verbose output for subprocess errors, we could catch `CalledProcessError` and exit more quietly.  Let me know if you'd like me to do so in this pull request.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -220,7 +220,7 @@ class UrlExtractor(HTMLParser):
 
 def runCmd(cmd):
     print(" ".join(cmd))
-    return subprocess.call(cmd)
+    subprocess.check_call(cmd)
 
 
 def execCmd(cmd, args):
@@ -318,8 +318,7 @@ def runJavac(sourceDir, classDir, classPath):
         args.append('--release')
         args.append(javaTargetVersion)
     args.append('@temp-javac-list')
-    if runCmd(args):
-        sys.exit(1)
+    runCmd(args)
     removeIfExists("temp-javac-list")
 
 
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:
> P.S, If you'd prefer that `build.py` produce less technical/verbose output for subprocess errors, we could catch `CalledProcessError` and exit more quietly. Let me know if you'd like me to do so in this pull request.

Nah, I think for our purposes, it’s just fine as-is.

Thanks for having put this together

kevinoid:
> Nah, I think for our purposes, it’s just fine as-is.

Sounds good to me.

> Thanks for having put this together

My pleasure.  Thanks for reviewing and merging it!

PR #1777, opened 2025-01-03
Rescue the Usage header in npm README

yecril25:
1. The Usage header should be emitted as a separate line.
2. Explain the intent in comments.


sideshowbarker:
In https://github.com/validator/validator/commit/6e062c7bc755252afd5dfdd73178e3b4ef60bac3 I made some changes to the npm packaging process that I think obviate the need for the change in this PR.

PR #1826, opened 2025-05-17
Don't create the DetectorFactory object if no language detection.

JustMisha:
The memory-consuming DetectorFactory object is only created if the system property "nu.validator.checker.enableLangDetection" is set to "1."


sideshowbarker (approved):

sideshowbarker:
Thanks

Issue #1569, opened 2023-03-23
<search> element


zcorpan:
URL of WHATWG spec or PR: https://github.com/web-platform-tests/wpt/pull/39163

**Describe the solution you'd like**
Validate the `search` element as per html and html-aria (https://github.com/w3c/html-aria/pull/401). Update the HTML parser.

mohamedatefka:
> URL of WHATWG spec or PR: https://github.com/web-platform-tests/wpt/pull/39163
> 
> **Describe the solution you'd like**
> Validate the `search` element as per html and html-aria (https://github.com/w3c/html-aria/pull/401). Update the HTML parser.
>

edent:
The validator currently shows an error when using `<section><search>`

> Element [search](https://html.spec.whatwg.org/multipage/#the-search-element) not allowed as child of element [section](https://html.spec.whatwg.org/multipage/#the-section-element) in this context. (Suppressing further errors from this subtree.)
> Contexts in which element [search](https://html.spec.whatwg.org/multipage/#the-search-element) may be used:
> ... Where [flow content](https://html.spec.whatwg.org/multipage/#flow-content-2) is expected.
> Content model for element [section](https://html.spec.whatwg.org/multipage/#the-section-element):
> ...  [Flow content](https://html.spec.whatwg.org/multipage/#flow-content-2).

Which seems rather contradictory!

(Search support has arrived in [Safari 17](https://webkit.org/blog/14445/webkit-features-in-safari-17-0/))

jribbens:
@edent Haha I just came here to report this because of your blog.

The following replicates the error - as far as I can see according to the HTML specification (as of right now) this is valid, but the validator complains.
```html
<!DOCTYPE html>
<html lang=en><meta charset=utf-8><title>Search test</title>
<section><h1>Search</h1><search>
<form><label>Search<input name="q"></label></form>
</search></section>
```

what1s1ove:
Hi everyone! Any news on this?

I'm facing an issue with my CI setup that involves validating .html files. Due to this error, I'm unable to start using a new tag 🥲

stephanboeni:
The example from https://html.spec.whatwg.org/multipage/grouping-content.html#the-search-element throws an error.

`<!DOCTYPE html>
<html lang="">
<head>
<title>Test</title>
</head>
<body>
<header>
  <h1><a href="/">My fancy blog</a></h1>
  ...
  <search>
    <form action="search.php">
      <label for="query">Find an article</label>
[...]

PR #1797, opened 2025-02-26
Add support for the <search> element.

EnergeticBark:
Closes #1569

Allows for the ARIA roles defined in: https://github.com/w3c/html-aria/pull/401

PR #1828, opened 2025-05-26
feat: add support for HTML `search` element

j9t:
Bear with me—on the deep end of the pool here. Got the validator to run locally and the proposed changes to validate the `search` element. Could not make out any other necessary changes, but am prepared to make additional adjustments.

(This would address #1569.)


j9t:
(Just noticed #1797. Can’t review now but will have a look. Happy to coordinate if this PR isn’t entirely useless 🙂)

PR #1833, opened 2025-06-07
use -Xss argument to run vnu.jar checks when stacksize is set

tenzap:
Otherwise, the invocation would fail on i386
```
  StackOverflowError while evaluating HTML schema.
  The checker requires a java thread stack size of at least 512k.
  Consider invoking java with the -Xss option. For example:
    java -Xss512k -jar ~/vnu.jar FILE.html
```
Using `_JAVA_OPTIONS `doesn't work because of https://stackoverflow.com/q/49233449/15401262


sideshowbarker:
Bump me after a week or so

PR #1836, opened 2025-06-07
Simplify way to get ImplementationVersion

tenzap:
While trying to create a Debian package for vnu for most dependencies I use the jars shipped by Debian. So I updated the build script to set the Class-Path in the Manifest.

However, this results in wrong or undefined value of the version. Apparently, when the Class-Path of the manifest references multiple jars, version will be updated by each iteration on the packages in the classpath, and the version results in being incorrect or unset.

This change will simplify the query and ensures that we get the version of the current package, without iterating over every dependency.


sideshowbarker:
Thanks — this looks like a good fix but I need to make time to actually test it myself before merging it. If I've not gotten around to merging it within the next week or so, please post a comment here to remind me

sideshowbarker (approved):

PR #1881, opened 2025-09-27
Make order of command-line options consistent

mernst:
The command-line options were listed in three different orders.
This pull request makes all three lists consistent.
It does not change the [Options](https://validator.github.io/validator/#options) section of the documentation, but changes the other two lists to be consistent with it.


sideshowbarker:
Michael, thanks much for taking time to contribute this — and sorry it took me so long to get around to merging it

PR #1897, opened 2025-10-20
Continue migration to ant-based build

tenzap:
With these changes, builds can be triggered completely by invoking only ant, and build.py would be here only for all other tasks not directly related to the production of the artifacts/jars/wars... That way build.py remains mostly a script to launch the binaries, distribute/package them, run the tests. However it still contains the code to produce the files in `nu/validator/localentities/files`.

The only thing I see which would be different between invoking a target directly from ant instead of calling the equivalent method in python is the "version" string. When running ant without setting the version property, it will use `nightly-${DSTAMP}`. When ant is invoked through build.py, the version is passed to ant by python. BTW, in some cases, build.py passes also the `dist` property. Maybe that could be improved and removed.

I removed the use of VERSION file in one of the commits. See comment in the commit message.

To have a clean tree, I started migration to have all artifacts built inside `build/<module>` instead of at the root of each module.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -146,3 +146,77 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  ant-build:
+    name: Ant build ${{ matrix.target }}
+    needs: prep
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target: [
+          build,
+          jar,
+          war,
+          cssvalidator-build,
+          cssvalidator-bundle,
+          galimatias-build,
+          galimatias-bundle,
+          htmlparser-build,
+          htmlparser-bundle,
+          jing-build,
+          jing-bundle,
+          langdetect-build,
+          langdetect-bundle,
+          validator-build,
+          validator-bundle,
+          validator-compare-full-and-merged-split,
+          validator-split-client,
+          validator-split-core,
+          validator-split-server,
+          ]
+    steps:
+      - uses: actions/checkout@v5
+      - name: Set up Java JDK
+        uses: actions/setup-java@v5
+        with:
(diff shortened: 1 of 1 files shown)


sideshowbarker:
There’s a merge conflict here now due to the change I made in https://github.com/validator/validator/commit/82f5a97a6a8ad05540964370192a1ee1e7156cc2 — and maybe in https://github.com/validator/validator/commit/a795ca1f42aa57482e4323a77d08a8f98d19a1f6.

The reason for those changes is: Just so that the filenames of the Maven pom file and Maven jar artifacts have the short version in their filenames — rather than the version that contains the hash in parentheses.

But I’m not wedded to the particular way I implemented that. Anything that achieves the same effect would be fine.

tenzap:
Hopefully I solved it without adding issues. Please note that you may have to proceed with the same change you did in validator also for `build/.ant/*.xml` because they also use the `${version}`.

And please update it after merging this one given the size of this PR.

sideshowbarker:
> Hopefully I solved it without adding issues. Please note that you may have to proceed with the same change you did in validator also for `build/.ant/*.xml` because they also use the `${version}`.

Unless for your (Debian packaging) purposes you need me to update those, then I think I won’t bother. As I recall, those are only used for creating Maven packages. And if so, since I’m no longer going to ever again be releasing new Maven packages for anything other than the validator itself, then there’d be no point in bothering to update those in any way.

tenzap:
What about removing that dead code then instead (maven related...)? Leaving dead code in place leads to loss of time to all (bugs, maintenance, questions...).

sideshowbarker:
> What about removing that dead code then instead (maven related...)? Leaving dead code in place leads to loss of time to all (bugs, maintenance, questions...).

Yeah I plan on getting around to yanking all that stuff out eventually

sideshowbarker:
> What about removing that dead code then instead (maven related...)? Leaving dead code in place leads to loss of time to all (bugs, maintenance, questions...).

For doing that, https://github.com/validator/validator/pull/1911 is the first half — the ant part. (I’ll do the python part after.)

PR #1898, opened 2025-10-20
Store all "validator" build artifacts in build/*

tenzap:
There shouldn't be any build product in the project/source tree anymore.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -44,6 +44,8 @@ DEALINGS IN THE SOFTWARE.
   <property name="validator-sources-jar" value="${dist}/${artifactId}-${version.number}-sources.jar" />
   <property name="validator-bundle-jar" value="${dist}/${artifactId}-${version.number}-bundle.jar" />
 
+  <!-- The default values of these 'validator.param.*' properties are supposed
+  to be equal to the values in build.py -->
   <property name="validator.param.aboutFile" value="site/about.html"/>
   <property name="validator.param.formTemplate" value="site/FormEmitter.xml"/>
   <property name="validator.param.pageTemplate" value="site/PageEmitter.xml"/>


sideshowbarker (approved):
Beautiful

sideshowbarker:
After these changes, does `java -jar ./build/dist/vnu.jar --version` work locally for you?

Or `./build/dist/vnu-runtime-image/bin/vnu --version`?

For me in my environment, both of those now longer emit the version as expected.

tenzap:
You've got to update the code to support also a "version" file that has no newline at the end of the value (which is what [echo](https://github.com/validator/validator/commit/7189cba963e1fd566e2f1d11f13fdb75db382ae0#diff-bffc59192cdccaf509753556461ff111ed11b8611616b7ebf7bbfd0700c3a711R849) does). Or get the version from another source (nu.validator.servlet.version in misc.properties (which is written by build.xml a few lines above)?, the manifest?, store it in a version.properties file instead ?...)

sideshowbarker:
> You've got to update the code to support also a "version" file that has no newline at the end of the value (which is what [echo](https://github.com/validator/validator/commit/7189cba963e1fd566e2f1d11f13fdb75db382ae0#diff-bffc59192cdccaf509753556461ff111ed11b8611616b7ebf7bbfd0700c3a711R849) does).

Yeah, I’ve subsequently discovered that. I guess the existing Java code is fragile to begin with it doesn’t work as expected for files that have no EOL/newline.

Still, I’m curious why ant’s `echo file` thing in that that `<echo file="${filesDir}/version">${version}</echo>` doesn’t already write the file with an newline.

Is it a lot of trouble to make that ant bit write the file with a newline?

If so, then yeah I can easily enough update the Java code to not fail if there’s no newline.

> Or get the version from another source (nu.validator.servlet.version in misc.properties (which is written by build.xml a few lines above)?, the manifest?, store it in a version.properties file instead ?...)

Yeah there may be some better place to get it from or to store it. But it can’t be the manifest at least — because the (jlink-generated) binary images aren’t packaged in any way with any manifest.

sideshowbarker:
For now just added a `flush()` that we rightly should have had there to begin with anyway

tenzap:
> Still, I’m curious why ant’s echo file thing in that that <echo file="${filesDir}/version">${version}</echo> doesn’t already write the file with an newline.

It's explained in the [doc](https://ant.apache.org/manual/Tasks/echo.html).

I think a property would be a good place to get the version from. Reading from a property doesn't require to implement the reading of a file.

With your change, on linux, the output is now:

```console
user@host:/home/user/validator$ java -jar build/dist/vnu.jar  --version
25.10.21user@host:/home/user/validator$ 
```

sideshowbarker:
> I think a property would be a good place to get the version from. Reading from a property doesn't require to implement the reading of a file.

Agreed, you’re right. I’ll switch it to read from a property. Which is anyway what we should’ve been doing from the beginning

> With your change, on linux, the output is now:
> 
> ```
> user@host:/home/user/validator$ java -jar build/dist/vnu.jar  --version
> 25.10.21user@host:/home/user/validator$ 
> ```

oops … it works in my macOS environment at least. But anyway the fact that the file-reading/writing code doesn’t behave consistently across platforms is yet another good reason to not being doing it that way.

> > Still, I’m curious why ant’s echo file thing in that that ${version} doesn’t already write the file with an newline.
> 
> It's explained in the [doc](https://ant.apache.org/manual/Tasks/echo.html).

I don’t actually find any explanation there that’s clear at all to me at least. Instead the only relevant thing I see is this:

> _The newline immediately following the `<echo>` tag is part of the output._

Specifically, I don’t see anything which makes it clear at all that the `<echo file="foo">bar</echo>` doesn’t write an EOL/EOF to the `foo` file.

In this particular case, it’s all moot — since I’m going to change the `--version` thing to not be writing and reading from a file anyway. But outside of this immediate particular case, it’s just seems surprising and suboptimal to me that Ant’s `<echo file="${filesDir}/version">${version}</echo>` thing doesn’t write EOL/EOF to file output — to the point that, if we’re using that in other places, it makes me think it could be a good idea to add something additional to any of those calls to ensure that the output files _do_ actually end up with EOL/EOF.

tenzap:
If we wanted a new line we would have to have it between the echo tags. That's how I understand it. Anyway, IMHO this is not the appropriate fix but just a dirty workaround.

tenzap:
I don't know where you saw that there is no manifest in the war. By reading the [ant doc of the `<war>` element](https://ant.apache.org/manual/Tasks/war.html), there is one built into the file at `META-INF/MANIFEST.MF`. Adding the implementation version, like it is done in the <jar> element does the trick.

Besides that, there used to be a simpler way to get the version, directly from the manifest, like in 52eb4aeab30c086f4af3f6cf5f3e79c61f26d457. What was wrong with this?

EDIT: sorry, you weren't talking about the war, but about the runtime image, so my comment is now partly irrelevant. However, isn't it possible to add one? The [ant doc of the `<jlink>` element](https://ant.apache.org/manual/Tasks/jlink.html) suggests that it would be possible. Maybe [that](https://stackoverflow.com/questions/61788739/what-is-the-equivalent-to-jar-file-meta-inf-manifest-mf-for-a-jimage-file#comment109441745_61802100) is also of interest?

EDIT2: Actually, I checked with the last released version 20.6.30 for which there are binaries (that's pretty old for a reference to catch regressions... but since this project doesn't make intermediate releases we'll have to do with it). With it, I found that version of the runtime image is properly set. Maybe 52eb4aeab30c086f4af3f6cf5f3e79c61f26d457, by fixing the issue for the jar, created a regression on the runtime-image... Because there is a manifest in the `module` of the runtime image, located inside the module at `vnu/META-INF/MANIFEST.MF`

sideshowbarker:
> Besides that, there used to be a simpler way to get the version, directly from the manifest, like in [52eb4ae](https://github.com/validator/validator/commit/52eb4aeab30c086f4af3f6cf5f3e79c61f26d457). What was wrong with this?

I don’t know now. Maybe nothing or maybe instead I found that it didn’t work.

Anyway, my intent at this point for ensuring we get the version back right across the jar/war/image versions is to go ahead with what you already suggested earlier — getting it from a system property. Unless you’re now suggesting that using `SimpleCommandLineValidator.class.getPackage().getImplementationVersion()` is better from some reason.

tenzap:
Hopefully you have read my latest version of my comment (given that I made multiple edits). I don't know what is better, I don't have much java knowledge. Only the very basics, dated 20 years back and never refreshed, at best.

sideshowbarker:
> Hopefully you have read my latest version of my comment (given that I made multiple edits). I don't know what is better, I don't have much java knowledge. Only the very basics, dated 20 years back and never refreshed, at best.

Yeah, given that I did some futzing several years ago with trying to `getImplementationVersion()` working reliably for the binary-image case and it ended up breaking — or I dunno maybe it never worked to begin but I just never noticed — I anyway don’t trust it at this point, and I reckon it’s safer to instead rely on something that we control at somewhat less of a level of abstraction. Which I guess it seems is best done by using a system property.

PR #1903, opened 2025-10-21
Permit to load html5spec from file (optionally)

tenzap:
Use case: when vnu is distributed in a distro (eg. Debian...), the version of vnu is fixed in a given release of the distro. Yet, the user may want to use vnu with a more recent html5spec copy that is located on his filesystem. It could for example be located in /var/lib/vnu/.

We could also imagine that the html5spec, which is an external resource to this project, is not distributed in the vnu package itself, but in another package. In such a setup, vnu could use the html5spec from that location more easily in place of the built-in html5spec. This could be configured through a /etc/vnu/vnu.properties file for example.

In this commit, the path to the html5spec is searched the vnu.properties file. Ideally, this would be searched in /etc/vnu/vnu.properties in case vnu is deployed on tomcat/jetty... The System property that was removed in 85cf3b13f143474c4db5a33905f64c46f4f27874 is now restored through a file.

Property name: nu.validator.spec.html5-load
Property value: path to the html5spec on the filesystem

Please think twice about it before merging.

### DIFF
--- a/src/nu/validator/localentities/LocalCacheEntityResolver.java
+++ b/src/nu/validator/localentities/LocalCacheEntityResolver.java
@@ -1,6 +1,8 @@
 package nu.validator.localentities;
 
 import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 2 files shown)


tenzap:
Actually I'm unsure if the property file in /etc/vnu would be accessible from a tomcat or jetty server
Maybe not for security reasons, but then I don't know how to easily configure this when vnu is installed on the system as a web app.

sideshowbarker:
> Please think twice about it before merging.

I’m actually not going to spend a lot of time thinking a lot about it or doing any hand-wringing about trying myself to anticipate any problems.

It’s an optional no-default thing, right? In the limited time I have to review, I care vastly more about what the default behavior is. With this, I realy don’t even know how many other users will even end up using it.

So for me, it’s sufficient just to know that this is something useful to you, and you’ve taken time to do the work of implementing it.

All that said, I’m also not going to put any time myself into maintaining this over the long run. So if we end up getting issues reported about it, and at that time you’re no longer around to do the work of triaging the issues and making any necessary fixes yourself, then likely what’s going to happen is that I’d just revert this.

sideshowbarker:
> Actually I'm unsure if the property file in /etc/vnu would be accessible from a tomcat or jetty server Maybe not for security reasons, but then I don't know how to easily configure this when vnu is installed on the system as a web app.

I don’t know either — and I’m not going to stop now and spend time myself testing this to try to figure that out.

If this doesn’t work in those cases, then it doesn’t work in those cases, and people won’t be able to use it in those cases but will instead need to use rely on the existing default behavior.

But for landing this, to me it’s sufficient that you do have a use case where you need it and you do know it works.

PR #1906, opened 2025-10-22
CLI: Get the version string from a Java system property

sideshowbarker:


sideshowbarker:
@tenzap please take a look

tenzap:
Should be ok

PR #1907, opened 2025-10-22
git: Ignore build/html5spec

sideshowbarker:


sideshowbarker:
@tenzap this seems to be needed now?

tenzap:
Yes probably. Thanks

tenzap:
Maybe it should be moved one line up (to keep new additions sorted)

sideshowbarker:
> Maybe it should be moved one line up (to keep new additions sorted)

OK yeah made that change

PR #1909, opened 2025-10-22
CI: Only use the full validator version in CI, not local builds

sideshowbarker:
@tenzap, please take a look

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,11 +40,13 @@ jobs:
       - name: Install dependencies
         run: python checker.py dldeps
       - name: Build
+        shell: bash
         run: |
-          python checker.py build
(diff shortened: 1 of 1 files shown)


tenzap:
I have added a `--version` option to build.py, maybe you should use that instead ? Or add a new option like `--version-suffix=<theSuffix>`. Adding new tasks to `build.py` just for this is too much duplication of code IMHO. Or `--use-git-version` That way you can keep the function in build.py?

sideshowbarker:
> I have added a `--version` option to build.py, maybe you should use that instead ?

Ah yeah, you’re absolutely right — just using that makes way more sense. I’ve pushed an update that makes it that way.

tenzap:
ok :)

PR #1910, opened 2025-10-22
ci: Use an environment variable for java version where appropriate

sideshowbarker:
@tenzap

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
 
+env:
+  JAVA_LATEST_VERSION: 25
+
 jobs:
   prep:
     name: Prepare the build
@@ -19,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        java: [25, 21, 17, 11.0.23]
+        java: [$JAVA_LATEST_VERSION, 21, 17, 11.0.23]
         exclude:
           - os: macos-latest
             java: 11.0.23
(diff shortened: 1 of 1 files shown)


tenzap:
Not here ? Line 22 or 25

java: [25, 21, 17, 11.0.23]

sideshowbarker:
> Not here ? Line 22 or 25
> 
> java: [25, 21, 17, 11.0.23]

Yeah, I suppose that makes sense.

tenzap:
Jobs are failing. Maybe it should be {{env.var_name}} or something like that.

sideshowbarker:
> Not here ? Line 22 or 25
> 
> java: [25, 21, 17, 11.0.23]

Apparently we can’t use an environment variable there. See https://github.com/validator/validator/actions/runs/18716709414/job/53377881607?pr=1912

It expects a literal, essentially.

sideshowbarker:
> Jobs are failing. Maybe it should be {{env.var_name}} or something like that.

I tried this:

```
matrix.java == ${{ env.JAVA_LATEST_VERSION }}
```

…but for that I get this error:

> matrix.java == ${{ env.JAVA_LATEST_VERSION }}\n" is always evaluated to true because extra characters are around ${{ }}

Any other ideas for the right syntax?

tenzap:
And ${{ left == env.variable }} ?

sideshowbarker:
> And ${{ left == env.variable }} ?

yeah I tried that already locally and my local syntax linter says that’s not going to work

sideshowbarker:
I’ve now entirely reverted this. If you want to experiment with something further yourself, maybe you can find something that works. But at this point I’m personally giving up on it myself.

PR #1911, opened 2025-10-22
Yank out maven stuff for dependencies, and move all targets up into build.xml

sideshowbarker:
@tenzap


.github/workflows/build.yml
```diff
@@ -157,15 +157,10 @@ jobs:
           jar,
           war,
           cssvalidator-build,
-          cssvalidator-bundle,
```
tenzap: I believe those should belong to the 1st commit, for consistency.

sideshowbarker: > I believe those should belong to the 1st commit, for consistency.

Yeah that makes sense. Made it so

tenzap:
ok

PR #1913, opened 2025-10-23
Support loading the language-profiles from local filesystem

tenzap:
Vnu will read 2 properties in the local vnu.properties files:

- nu.validator.checker.language-profiles-list
- nu.validator.checker.language-profiles-dir

These are the path on the filesystem to a local copy of the language-profiles dir, and of the language-profiles-list.txt file.

If these properties are set, vnu will use them instead of the built-in version.

This can be useful if one wants to use alternative/refreshed language detection profiles (but still valid for the langdetect library).

### DIFF
--- a/src/nu/validator/checker/LanguageDetectingChecker.java
+++ b/src/nu/validator/checker/LanguageDetectingChecker.java
@@ -33,15 +33,21 @@ import io.mola.galimatias.Host;
 import io.mola.galimatias.URL;
 
 import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:
So now every single time somebody runs the CLI, it emits the following to stdout?

> nu.validator.checker.LanguageDetectingChecker - Optional vnu.properties file not found.

Is that intentional?

sideshowbarker:
I’ve reverted this for now.

Our code should not be causing the CLI to emitting a message like _“Optional vnu.properties file not found.”_ to stdout every single time it’s run. And it shouldn’t be emitting anything informational unless the user has set the `verbose` option.

It also probably shouldn’t use log4j just to emit messages like this.

PR #1924, opened 2025-10-25
build.py: add task to build specific target with ant

tenzap:


sideshowbarker (approved):
Lovely stuff

sideshowbarker:
I didn't know about `shlex` until now. It's great and I guess we should be using it everywhere we can.

PR #1964, opened 2025-12-05
Add option to only show HTML errors and warnings

### LINKED ISSUE #940: Please allow to disable CSS validation (distracts from HTML validation)
In order to focus only on HTML validation it would be useful to have an option to avoid (disable) any kind of CSS validation, even for inline and internal CSS, or, which seems suitable in case of an HTML validator, to have an option to opt _into_ CSS validation rather than needing to opt out of it.

Rationale: It seems both counter-intuitive and counter-productive to get reports on CSS issues when using an HTML validator for HTML validation—especially when reports on CSS issues end up making it difficult to find HTML issues. (I just had this problem with a local test file that contained perfectly valid HTML and CSS, but that triggered a number of false positives with the internal CSS. That made it surprisingly difficult to tell the status of the HTML. The problem here was both the unexpected validation of CSS and then the problem that the CSS validation was faulty.)

Please let me know if I’m missing something—#621 contained some hint at an option to disable all CSS validation, but it was surprising just now that the HTML validator could so badly stand in its own way 😬 Thanks!

j9t:
Made a simple adjustment to allow to show and reproduce only HTML issues (available under “Message Filtering,” then “show only HTML errors” and “show only HTML warnings”).

@sideshowbarker, tried to test locally but seem to end up on an interface I’m not familiar with:

Screenshot of local validator.

Poked around but couldn’t get to the version providing message filtering. Can you advise or test on your end?

Resolves #940

### DIFF
--- a/site/nu-script.js
+++ b/site/nu-script.js
@@ -571,7 +571,7 @@ function updateFragmentIdHilite() {
 			newStyle.styleSheet.cssText = rule
 		} else {
 			throw ex
-		} 
+		}
 	}
 	dynamicStyle.parentNode.replaceChild(newStyle, dynamicStyle)
 	dynamicStyle = newStyle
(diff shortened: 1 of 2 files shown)


sideshowbarker:
1. Run this:
   ```
   ./checker.py clean build test run
   ```
2. Open http://localhost:8888 in your browser.

j9t:
@sideshowbarker, thanks! I had just been running `python ./checker.py run`.

Got to test and everything works well now. The change would allow to filter and show only HTML errors, warnings, and messages (and their counts), _if_ there are messages related to anything non-HTML.

PR is updated and ready for review 🙂

sideshowbarker:
Can you give me a URL or document to test with?

j9t:
> Can you give me a URL or document to test with?

Sure—here’s a test file that would show mixed errors and warnings: [test.html](https://github.com/user-attachments/files/23981763/test.html)

This would then lead to a view like this:

Screenshot of new “show only HTML…” message feature.

sideshowbarker (approved):

PR #1968, opened 2025-12-09
fix: handle wildcard patterns in element and attribute checks

### LINKED ISSUE #729: Internal error when validating XML
Hi!

I believe, I have found a bug. Trying to validate the following XML, results in "_Oops. That was not supposed to happen. A bug manifested itself in the application internals. Unable to continue. Sorry. The admin was notified._" The RNG schema specifies that the element "imagedata" must have a mandatory "fileref" attribute. So, after adding that attribute, the document validates and the error disappears.

The (invalid) XML document looks like this:
`<?xml version="1.0" encoding="utf-8" ?>
<section xmlns="http://docbook.org/ns/docbook" version="5.0">
  <title></title>
  <para></para>
  <mediaobject>
    <imageobject>
      <imagedata/>
    </imageobject>
  </mediaobject>
</section>`

And the schema can be found here: https://docbook.org/xml/5.0.1/rng/docbook.rng

I was not able to anything useful on Google this time, I am afraid. Is this a known issue?
Thanks for an outstanding piece of software!

j9t:
A bit at the deeper end of the pool here but this should fix #729. By skipping wildcard patterns, the validator should continue instead of crashing.

Closes #729

### DIFF
--- a/jing-trang/mod/rng-validate/src/main/com/thaiopensource/relaxng/impl/PatternValidator.java
+++ b/jing-trang/mod/rng-validate/src/main/com/thaiopensource/relaxng/impl/PatternValidator.java
@@ -224,26 +224,34 @@ public class PatternValidator extends DtdContext implements Validator, ContentHa
 
     /**
      * Got an element pattern.
-     * 
+     *
      * @see com.thaiopensource.relaxng.impl.PatternVisitor#visitElement(com.thaiopensource.relaxng.impl.NameClass,
      *      com.thaiopensource.relaxng.impl.Pattern)
      */
     public void visitElement(NameClass nc, Pattern content) {
       // just output the element name class, not the content.
+      nameRecord = null;
       nc.accept(this);
-      elements.add(nameRecord);
+      // Only add if nameRecord was set (i.e., not a wildcard like anyName or nsName)
+      if (nameRecord != null) {
+        elements.add(nameRecord);
+      }
     }
 
     /**
      * Got an attribute pattern. Here it seems we get only attribute
      * wildcards patterns.
-     * 
+     *
      * @see com.thaiopensource.relaxng.impl.PatternVisitor#visitAttribute(com.thaiopensource.relaxng.impl.NameClass,
      *      com.thaiopensource.relaxng.impl.Pattern)
      */
     public void visitAttribute(NameClass ns, Pattern value) {
+      nameRecord = null;
       ns.accept(this);
-      attributes.add(nameRecord);
+      // Only add if nameRecord was set (i.e., not a wildcard like anyName or nsName)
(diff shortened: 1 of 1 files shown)


jing-trang/mod/rng-validate/src/main/com/thaiopensource/relaxng/impl/PatternValidator.java
```diff
@@ -224,26 +224,34 @@ public void visitOneOrMore(Pattern p) {
...
     public void visitElement(NameClass nc, Pattern content) {
       // just output the element name class, not the content.
+      nameRecord = null;
       nc.accept(this);
-      elements.add(nameRecord);
+      // Only add if nameRecord was set (i.e., not a wildcard like anyName or nsName)
+      if (nameRecord != null) {
+        elements.add(nameRecord);
+      }
     }
```
sideshowbarker: I don’t know what this code is actually doing.

I mean, it see that:

1. The `NameRecordingVisitor` class that this is code is in has a `nameRecord` data member (or “field” or whatever the term is for it in Java).

2. This  code sets that `nameRecord` data member to null — for… some reason. 

3. Then two lines later, this code checks to see if `nameRecord` is null — which it will of course be, _unless_ the `nc.accept(this)` call ends up setting it to something else.

   So I guess I can only assume it’s possible for `nc.accept(this)` to cause that `nameRecord` data member to be set to something other than null.
   
But in the end, none of that really matters — because I have to admit that at this point, I don’t actually even have a super-clear understanding of what we need this `visitElement()` function for to begin with.

While it’s true that I _have_ actually touched that code at one time, it was _15 years ago_ — and all I did with it was essentially just to cargo-cult copy over a change that another jing-trang contributor had made.

sideshowbarker: Yeah, I guess it’s best that we close this.

My single biggest concern about this change is, if it could inadvertently lead to some (possibly surprising and hard-to-debug) regression. And we don’t really have any existing regressions tests as such for this code.

In an ideal world, I would make time to go back and really study the code. And write some tests.

But as things stand now: Unless there’s some problem in this (jing) code that’s leading to a reproducible problem in checking HTML documents or SVG content, I am loathe to go in and start messing around it with.

And #729 is neither (1) reproducible (at least not from what I can glean from going back and re-reading the issue description and comments, nor (2) a problem that’s been reported with checking HTML documents or SVG.

Instead, it’s a problem with checking some content against an arbitrary schema (which happens to be a DocBook schema, but that doesn’t really matter). And checking content against any arbitrary (non-HTML or non-SVG) schema is not something we currently really support doing with the checker any longer anyway.

PR #1971, opened 2025-12-12
Persist input mode selection

### LINKED ISSUE #160: Web front-end should recall user's last-selected input type
I'm finding myself repeatedly changing the input mode to `text input` every time I open the web front-end.

Is it possible to set a cookie that remembers what the user last chose as their input type?

Happen to use this thing a lot during local development, and validating content by its URL isn't going to help if it's hosted locally.

j9t:
Uses local storage to persist input mode. (Includes minor consistency improvements to indentation and formatting.)

Resolves #160


sideshowbarker:
Nice!

Issue #1470, opened 2022-10-31
Feature request; help wanted: make the npm autoinstall Java


uuf6429:
It's very weird that the npm module just contains the jar. One would expect that installing a dependency through a dependency manager automatically resolves all dependencies.
In this case, AFAIK it doesn't even check for Java.

Installing the module on a java-less system succeeds without issues, _which is in principle, **very wrong**_.

The cheapest option is to check for java during install-time, but I don't know if or how that could be done in package.json (~~[I know it can be done for node](https://stackoverflow.com/a/29349773/314056), not sure about java~~ I guess this check could also be done in the install script, mentioned below).

The nicest option is to download the right binary for the current system, similar to how [phantomjs](https://github.com/Medium/phantomjs/blob/master/install.js#L403) does it (and perhaps default to the jar version when java is indeed installed).

At the moment this module is as useful to me as the download link in GitHub.

uuf6429:
This could be a potential solution: https://github.com/nvuillam/node-java-caller/tree/master/examples/cli_app

sideshowbarker:
The suggestions in the issue description sound great to me. But the npm package was put together by another contributor, and I don’t personally plan on ever doing any more work on it myself. And I don’t use the npm package for anything myself.

Therefore, if any of those suggestions are to ever actually be implemented, we’d need a patch/PR for it.

So, I’ve:

- labeled this issue with the “help wanted” label
- pinned it to the top of the issues list
- updated the title to reflect what specific changes it suggests

 I know that may not mean much, but it’s all I can offer.

Edgarruiz856:
- [x]  edgarplasticos3@hotmail.com> It's very weird that the npm module just contains the jar. One would expect that installing a dependency through a dependency manager automatically resolves all dependencies.
> In this case, AFAIK it doesn't even check for Java.
> 
> Installing the module on a java-less system succeeds without issues, _which is in principle, **very wrong**_.
> 
> The cheapest option is to check for java during install-time, but I don't know if or how that could be done in package.json (~~[I know it can be done for node](https://stackoverflow.com/a/29349773/314056), not sure about java~~ I guess this check could also be done in the install script, mentioned below).
> 
> The nicest option is to download the right binary for the current system, similar to how [phantomjs](https://github.com/Medium/phantomjs/blob/master/install.js#L403) does it (and perhaps default to the jar version when java is indeed installed).
> 
> At the moment this module is as useful to me as the download link in GitHub.

PR #1972, opened 2025-12-13
feat: Make the npm package optionally autoinstall Java

sideshowbarker:
Fixes https://github.com/validator/validator/issues/1470. Fixes https://github.com/validator/validator/issues/1974.

### DIFF
--- a/vnu-jar.js
+++ b/vnu-jar.js
@@ -1,3 +1,39 @@
+#!/usr/bin/env node
 'use strict';
+const { execFile } = require('child_process');
+const path = require('path');
+const vnuJar = path.join(__dirname, 'build/dist/vnu.jar');
+const vnu = {
+    check: async function (args = [], options = {}) {
+        const { resolveJava } = require('./vnu-java-downloader');
+        const javaCmd = await resolveJava();
+        const finalArgs = ['-jar', vnuJar, ...args];
+        return new Promise((resolve, reject) => {
+            const child = execFile(javaCmd, finalArgs, (error, stdout, stderr) => {
+                const output = [stdout, stderr].filter(Boolean).join('');
+                if (error) {
+                    const errObj = new Error(output || error.message);
+                    reject(errObj);
+                } else {
+                    resolve(output);
+                }
+            });
+        });
+    }
+};
 
-module.exports = require('path').join( __dirname, 'build/dist/vnu.jar' );
+const exported = new String(vnuJar);
+exported.vnu = vnu;
+module.exports = exported;
+
+if (require.main === module) {
+    (async () => {
+        try {
+            console.log(await vnu.check(process.argv.slice(2)));
+        } catch (err) {
+            console.error(err.message.trim());
+            process.exit(1);
+        }
+    })();
+}
+
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,5 @@ extras/
 jars.tar.gz
 jars/
 local-entities/
+node_modules/
 run-validator.sh
--- a/npm.md
+++ b/npm.md
@@ -8,30 +8,34 @@
 npm install --save vnu-jar
 ```
 
-## Install 'next' version
-
-```sh
-npm install --save vnu-jar@next
-```
-
 ## Example
+
+Here’s an example of how to create a simple `vnu-check.js` program that uses the `vnu-jar` package to invoke the Nu Html Checker (vnu), pass it whatever arguments the user specifies, and then emit the checker output.
 
 ```js
 'use strict';
-
-const { execFile } = require('child_process');
-const vnu = require('vnu-jar');
-
-// Print path to vnu.jar
-console.log(vnu);
-
-// Work with vnu.jar, for example get vnu.jar version
-execFile('java', ['-jar', `"${vnu}"`, '--version'], { shell: true }, (error, stdout) => {
-    if (error) {
-        console.error(`exec error: ${error}`);
-        return;
+const { vnu } = require('vnu-jar');
+(async () => {
+    try {
+        console.log(await vnu.check(process.argv.slice(2)));
+    } catch (err) {
+        console.error(err.message.trim());
+        process.exit(1);
     }
-
-    console.log(stdout);
-});
+})();
 ```
+
+## Command-line usage
+
+When installed or run via `npx`, the package also provides a `vnu` command that causes the `vnu.jar` program to be invoked and run with whatever arguments you specify:
+
+```sh
+npx vnu --verbose file1.html file2.html
+npx vnu --help
+```
+
+## Java auto-installation
+
+This npm package includes the `vnu.jar` Java program — so to use the package, users will additionally need a Java environment that provides a `java` command.
+
+For that reason, the package runs a `postinstall` script which — if it finds no `java` command in the user’s environment during install — will then _automatically_ install a (Node.js-local) Java runtime environment, in `node_modules/.cache/vnu-jar/java/`).
--- a/package.json
+++ b/package.json
@@ -29,6 +29,13 @@
   ],
   "files": [
     "build/dist/vnu.jar",
+    "vnu-java-downloader.js",
     "vnu-jar.js"
-  ]
+  ],
+  "bin": {
+    "vnu": "./vnu-jar.js"
+  },
+  "scripts": {
+    "postinstall": "node vnu-java-downloader.js"
+  }
 }
--- a/vnu-java-downloader.js
+++ b/vnu-java-downloader.js
@@ -1,0 +1,110 @@
+#!/usr/bin/env node
+'use strict';
+
+const { spawnSync, execSync } = require('child_process');
+const { mkdirSync, existsSync, createWriteStream } = require('fs');
+const { join, dirname, parse } = require('path');
+const { Readable } = require('stream');
+
+function findNearestNodeModules(startDir) {
+    let dir = startDir;
+    const { root } = parse(startDir);
+    while (true) {
+        const candidate = join(dir, 'node_modules');
+        if (existsSync(candidate)) {
+            return candidate;
+        }
+        if (dir === root) {
+            throw new Error('Could not find a node_modules directory.');
+        }
+        dir = dirname(dir);
+    }
+}
+
+const NODE_MODULES_DIR = findNearestNodeModules(__dirname);
+const CACHE_DIR = join( NODE_MODULES_DIR, '.cache', 'vnu-jar', 'java');
+const TEMURIN_VERSION= '17.0.17+10'
+const TEMURIN_BASE_URL = `https://github.com/adoptium/temurin17-binaries/releases/download/jdk-${TEMURIN_VERSION}/`;
+
+function findSystemJava() {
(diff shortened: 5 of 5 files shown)


sideshowbarker:
> It looks fine from my perspective, but keep in mind I haven't touched this project in years

Thanks for having taken time to review, and for the fixes and suggested improvements — much appreciated.

> so it would be best to have at least one other person invested in this project to give it a look.

👍 Yeah — and I’ll at least try to make sure this isn’t going to break anything for anybody.

sideshowbarker:
@sidvishnoi: It would be great if you also could take a look at this one — if/when you have time.

vnu-java-resolver.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+const { spawnSync, execSync } = require('child_process');
+const { mkdirSync, existsSync, createWriteStream } = require('fs');
+const { join } = require('path');
+const https = require('https');
+const { extract } = require('tar');
+
+const CACHE_DIR = join(__dirname, '..', 'node_modules', '.cache', 'vnu-jar', 'java');
+const TEMURIN_BASE_URL = 'https://github.com/adoptium/temurin17-binaries/releases/latest/download/';
```
sidvishnoi: Should we consider locking to particular versions? In the event the latest release doesn't include a particular platform/arch build?

vnu-java-resolver.js
```diff
@@ -0,0 +1,90 @@
...
+    const whichCmd = process.platform === 'win32' ? 'where' : 'which';
+    const result = spawnSync(whichCmd, ['java'], { encoding: 'utf8' });
+    if (result.status === 0 && result.stdout) {
+        return result.stdout.split(/\r?\n/)[0].trim();
+    }
+    return null;
+}
+
+function getPlatformArchiveName() {
+    const arch = process.arch === 'x64' ? 'x64' : process.arch;
```
sidvishnoi: This is equivalent:
```suggestion
    const arch = process.arch;
```

vnu-java-resolver.js
```diff
@@ -0,0 +1,90 @@
...
+    await new Promise((resolve, reject) => {
+        const file = createWriteStream(archivePath);
+        https.get(url, res => {
+            if (res.statusCode !== 200) {
+                reject(new Error(`Failed to download Java: ${res.statusCode}`));
+                return;
+            }
+            res.pipe(file);
+            file.on('finish', () => file.close(resolve));
+        }).on('error', reject);
```
sidvishnoi: I'm not sure which Node.js version this package aims to target, but maybe using `fetch` might be better?

vnu-java-resolver.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+if (require.main === module) {
+    (async () => {
+        try {
+            await resolveJava();
+        } catch {
+            process.exit(0);
+        }
+    })();
```
sidvishnoi: Nit:
```suggestion
    void resolveJava().catch(() => process.exit(0));
```

vnu-java-resolver.js
sideshowbarker: > I'm not sure which Node.js version this package aims to target, but maybe using `fetch` might be better?

Thanks, yeah — should have used `fetch` to begin with…

vnu-java-resolver.js
```diff
@@ -0,0 +1,100 @@
...
+            return candidate;
+        }
+        if (dir === root) {
+            throw new Error('Could not find a node_modules directory.');
+        }
+        dir = dirname(dir);
+    }
+}
+
+const NODE_MODULES_DIR = findNearestNodeModules(process.cwd());
```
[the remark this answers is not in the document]
uuf6429: Why would it be safer? It's kinda clearer to users that an executable sees things relative to the current working directory, but a constant based on the location of some script feels arbitrary and unsafe. For instance, if the code or the script is moved, it would break. 🤔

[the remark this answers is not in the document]
sidvishnoi: I should've been clearer when I said _safer_. Consider running `vnu-java-resolver` like this:
```bash
$ ls
# library-that-uses-vnu-jar another-project

$ cd another-project

$ node ../library-that-uses-vnu-jar/index.js
# When resolving java, process.cwd() will point to `another-project`, 
# but the node_modules actually are in `library-that-uses-vnu-jar/node_modules`
# so the process won't find node_modules (as can be different dir at install & run time).
```

This is how [spec-prod](https://github.com/w3c/spec-prod) usually runs. Although it has java available globally in CI, local dev setups might actually use the logic in this PR.

nschonni (commented):
Having previously maintained node-sass that had a custom installer system/post-install scripts, I'm going to suggest this is a path you don't want to go down. It's complicated and will attract a new class of Issues in the repo. I think some sniffing for a missing Java install for those that use the binary make sense, but anything more, and you probably want to be pointing over to using a different package in the instructions. https://www.npmjs.com/package/java and https://www.npmjs.com/package/njre are 2 that I just found from search, not things I've actually used.

sideshowbarker:
> will attract a new class of Issues in the repo

That thought had occurred to me. So yeah, it would be good to avoid that.

> Having previously maintained node-sass that had a custom installer system/post-install scripts, I'm going to suggest this is a path you don't want to go down.

I hear you, and I understand — but I got to say that unless you strongly object, I’d like for us to go ahead with maybe  least trying this, and seeing what happens. I mean, it doesn’t seem it can do real damage to users, or be outright harmful — while on the other hand, it does seem that some could benefit from it. So I guess what I’m saying is, we could get it out there — and then, if/when we do end up getting a lot of issues about it, then I would be fine with just ruthlessly reverting it.

> I think some sniffing for a missing Java install for those that use the binary make sense, but anything more, and you probably want to be pointing over to using a different package in the instructions. [npmjs.com/package/java](https://www.npmjs.com/package/java) and [npmjs.com/package/njre](https://www.npmjs.com/package/njre) are 2 that I just found from search, not things I've actually used.

My first thought it is to wonder if we could use one of those as an actual dependency — that is, make it automatically handle the install in the same way as the ad-hoc thing I put together here.

But regardless, I guess that in making your suggestion, that’s not what you had in mind, anyway? I mean, what you’re suggesting is that we’d instead just put something about it into the package docs?

vnu-java-resolver.js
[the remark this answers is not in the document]
sideshowbarker: > I'm not sure if it's better to start from `__dirname` or `process.cwd()`. `__dirname` feels safer.

I have to admit I really don’t have enough experience with Node to have any opinion on that either way. But spec-prod is a key target for this package — and you certainly have way more insight on this — so if you tell me that on balance, it seems better to use `__dirname`, that works fine for me. So, shall I change it (back) to that?

vnu-java-resolver.js
```diff
@@ -0,0 +1,90 @@
...
+            res.pipe(file);
+            file.on('finish', () => file.close(resolve));
+        }).on('error', reject);
+    });
+    if (process.platform === 'win32') {
+        console.log('Extracting ZIP archive...');
+        execSync(`powershell -Command "Expand-Archive -Force '${archivePath}' '${CACHE_DIR}'"`);
+    } else {
+        console.log('Extracting tar.gz archive...');
+        await extract({ file: archivePath, cwd: CACHE_DIR, strip: 1 });
```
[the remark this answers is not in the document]
sideshowbarker: > I meant as adding the `tar` dependency from npm.

Ah yeah, I’d forgotten about that. Or well, I kinda didn’t understand that to begin with. I mean, I used `extract` without realizing that introduced a dependency on `tar`. But I guess it does? At least I can say that when tried it initially, it failed with an error message about a missing `tar` dependency. So that’s why I added the dependency.

So, anyway, to get back to your original question:

> Following consistency with Windows/Powershell, can we assume existence of `tar` on Linux/Mac?
 
I can only respond by saying: I dunno — and with a question of my own, which is: What kind of code changes would/should I make in response to that concern?

vnu-java-resolver.js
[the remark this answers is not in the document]
sideshowbarker: > Yes, let's use __dirname as the starting point.

👍 Made it so.

vnu-java-resolver.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
+
+const { spawnSync, execSync } = require('child_process');
+const { mkdirSync, existsSync, createWriteStream } = require('fs');
+const { join, dirname, parse } = require('path');
+const { extract } = require('tar');
```
sidvishnoi: Can remove this dependency and import now.

vnu-jar.js
```diff
@@ -1,3 +1,19 @@
 'use strict';
-
-module.exports = require('path').join( __dirname, 'build/dist/vnu.jar' );
+const { execFile } = require('child_process');
+const path = require('path');
+const { resolveJava } = require('./vnu-java-resolver');
```
sidvishnoi: Consider importing this within `check`, so all the path traversals, JS imports can be skipped for the non-`check` use cases.

vnu-java-resolver.js
```diff
@@ -0,0 +1,100 @@
...
+async function downloadJava() {
+    mkdirSync(CACHE_DIR, { recursive: true });
+    const archiveName = getPlatformArchiveName();
+    const url = TEMURIN_BASE_URL + archiveName;
+    const archivePath = join(CACHE_DIR, archiveName);
+    console.log(`Downloading Java 17 runtime from ${url}...`);
+    const res = await fetch(url);
+    if (!res.ok) {
+        throw new Error(`Failed to download Java: ${res.status} ${res.statusText}`);
+    }
```
sidvishnoi: On my Macbook M1 Air, I got this error:
```
Downloading Java 17 runtime from https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.17%2B10/OpenJDK17U-jre_arm64_mac_hotspot_17.0.17_10.tar.gz...
Error: Failed to download Java: 404 Not Found
```

archiveName: `OpenJDK17U-jre_arm64_mac_hotspot_17.0.17_10.tar.gz`

vnu-java-resolver.js
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function getPlatformArchiveName() {
+    let arch = process.arch;
+    const plat = process.platform === 'win32' ? 'windows' : process.platform === 'darwin' ? 'mac' : 'linux';
+    if (plat === 'mac' && arch === 'arm64') {
+        arch = 'aarch64'
+    }
+    const ext = process.platform === 'win32' ? 'zip' : 'tar.gz';
+    return `OpenJDK17U-jre_${arch}_${plat}_hotspot_latest.${ext}`;
```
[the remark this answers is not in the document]
sideshowbarker: > Best extract the version to a variable, and use it in BASE_URL as well.

Yes. Made it so.

sideshowbarker:
Sorry for all the thrashing on this, but I think I may finally now have managed to get every thing right.

I really am not npm-savvy, and I have to confess that, until today when I actually looked into how it can be done, I didn’t even know how to “locally” install and test an npm package. But today I did finally get around to figuring that out, and then was able to test all this and make sure myself that I could get it all working as expected.

PR #1973, opened 2025-12-15
Extend warning for multiple `h1` elements (without offsets)

### LINKED ISSUE #813: Error in detect mutiple h1 elements
Multiple h1 in body dont throw error

### Example

```html
<!doctype html>
<html lang="en">
<head>
	<meta charset="utf-8">
	<title>title</title>
</head>
<body>
	<h1>one</h1>
	<h1>two</h1>
</body>
```

### Expected behavior

Warn or error about mutiple use **h1** tag

### Actual behavior

No errors

### Step to reproduce

Direct input or upload file with example on https://validator.w3.org/

j9t:
Improved handling of multiple `h1` elements, even taking into account `headingoffset`.

Resolves #813

### DIFF
--- a/src/nu/validator/checker/schematronequiv/Assertions.java
+++ b/src/nu/validator/checker/schematronequiv/Assertions.java
@@ -847,6 +847,12 @@ public class Assertions extends Checker {
             + " \u201Ch1\u201D elements are treated as top-level headings"
             + " by many screen readers and other tools).";
 
+    private static final String multipleH1WarningMessage = "Consider using"
+            + " only one \u201Ch1\u201D element per document (or, if using"
+            + " \u201Ch1\u201D elements multiple times is required, consider"
+            + " using the \u201Cheadingoffset\u201D attribute to indicate that"
+            + " these \u201Ch1\u201D elements are not all top-level headings).";
+
     private class IdrefLocator {
         private final Locator locator;
 
@@ -1542,7 +1548,7 @@ public class Assertions extends Checker {
 
     private int numberOfSvgAelementsDeep = 0;
 
-    private Set<Locator> secondLevelH1s = new HashSet<>();
+    private Set<Locator> allH1s = new HashSet<>();
 
     private Map<Locator, Map<String, String>> siblingSources = new ConcurrentHashMap<>();
 
@@ -1730,9 +1736,10 @@ public class Assertions extends Checker {
             }
         }
 
-        if (hasTopLevelH1 && !hasHeadingoffset) {
-            for (Locator locator : secondLevelH1s) {
-                warn(h1WarningMessage, locator);
+        // Warn if multiple h1 elements found in document
+        if (allH1s.size() > 1 && !hasHeadingoffset) {
+            for (Locator locator : allH1s) {
+                warn(multipleH1WarningMessage, locator);
             }
         }
 
@@ -2155,7 +2162,7 @@ public class Assertions extends Checker {
         tabpanelElements.clear();
         tabElementsActive.clear();
         siblingSources.clear();
-        secondLevelH1s.clear();
+        allH1s.clear();
         sectioningElementPtrs.clear();
     }
 
@@ -3104,10 +3111,12 @@ public class Assertions extends Checker {
                     hasVisibleMainRole = true; // <main> has implicit role="main"
                 }
             } else if ("h1" == localName && !hasHeadingoffset) {
+                // Store location of all h1 elements for multiple h1 check
+                allH1s.add(new LocatorImpl(getDocumentLocator()));
+
+                // Keep existing warning for deeply nested h1 elements (2+ levels)
                 if (sectioningElementPtrs.size() > 1) {
                     warn(h1WarningMessage);
-                } else if (sectioningElementPtrs.size() == 1) {
-                    secondLevelH1s.add(new LocatorImpl(getDocumentLocator()));
                 } else {
                     hasTopLevelH1 = true;
                 }
--- a/tests/html/elements/h1/multiple-h1-haswarn.html
+++ b/tests/html/elements/h1/multiple-h1-haswarn.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Multiple h1 elements</title>
+</head>
+<body>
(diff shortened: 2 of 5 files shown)


j9t:
@sideshowbarker, does this reflect the new expected process for tests? I noticed that https://github.com/validator/tests is archived even though it’s still mentioned in [.gitsubtrees.yaml](https://github.com/validator/validator/blob/main/.gitsubtrees.yaml).

sideshowbarker:
> @sideshowbarker, does this reflect the new expected process for tests?

Yeah — in that the tests should be added in the `tests` subdirectory.



> I noticed that [validator/tests](https://github.com/validator/tests?rgh-link-date=2025-12-15T10%3A03%3A37Z) is archived even though it’s still mentioned in [`.gitsubtrees.yaml`](https://github.com/validator/validator/blob/main/.gitsubtrees.yaml?rgh-link-date=2025-12-15T10%3A03%3A37Z).

Yeah, that’s all intentional. The original upstream repo which the tests came from is now read-only. And the inclusion in  `.gitsubtrees.yaml` is intentional — but only for the purpose of keeping a record of where the tests originally came from.

sideshowbarker:
One thing I had already been meaning to mention to you: Please see https://github.com/validator/validator/blob/main/CONTRIBUTING.md#commit-hooks for some instructions on how you can set up your local environment so that some git “pre-commit” checks in your local environment every time you commit.

If you got through those steps, then among the checks that will do is, it’ll check to make sure you’ve also updated the `tests/messages.json` file. Which you didn’t do yet in the patch for this PR — but you can do it by running `./checker.py make-messages`, after you’ve added any tests.

And in the case of this particular PR, what that will specifically do is, it’ll add a message there for the `tests/html/elements/h1/multiple-h1-haswarn.html` test.

j9t:
> One thing I had already been meaning to mention to you: Please see https://github.com/validator/validator/blob/main/CONTRIBUTING.md#commit-hooks for some instructions on how you can set up your local environment so that some git “pre-commit” checks in your local environment every time you commit.

Set up.

> If you got through those steps, then among the checks that will do is, it’ll check to make sure you’ve also updated the `tests/messages.json` file. Which you didn’t do yet in the patch for this PR — but you can do it by running `./checker.py make-messages`, after you’ve added any tests.

~[…]~ Seems to have an effect now—yes, set up!

sideshowbarker (approved):

sideshowbarker:
This seems to have broken some existing tests:

```
"file:/Users/mike/workspace/validator/tests/html/attributes/headingoffset-headingreset-isvalid.html":3:4: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/attributes/headingoffset-headingreset-isvalid.html":13:19: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/attributes/popover-isvalid.html":5:4: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/attributes/popover-isvalid.html":21:21: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/elements/h1/model-isvalid.html":15:6: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/elements/h1/model-isvalid.html":12:6: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
"file:/Users/mike/workspace/validator/tests/html/elements/h1/model-isvalid.html":9:38: warning: Consider using only one “h1” element per document (or, if using “h1” elements multiple times is required, consider using the “headingoffset” attribute to indicate that these “h1” elements are not all top-level headings).
```

sideshowbarker:
So, either those existing tests need to be renamed to `-haswarn.html` or else they need to fixed in some way, or else the change in this PR needs to be further refined.

Anyway, in the meantime I’m reverting this change — until we get the problems figured out and fixed.

j9t:
@sideshowbarker, what happened, why did local tests and CI pass?

Your call on that but is there or should there be a ticket to make sure these tests cannot be overlooked?

And, should we reopen https://github.com/validator/validator/issues/813?

sideshowbarker:
> @sideshowbarker, what happened, why did local tests and CI pass?

What happened is that this change was causing new _warnings_ to be emitted for `-valid.html` — but that didn’t cause CI to fail. And that’s because the test runner, as currently written, doesn’t treat warnings for `-valid.html` files as as failure. The test runner still exits zero in that case. So that’s why CI didn’t catch it, and we didn’t catch it.

> Your call on that but is there or should there be a ticket to make sure these tests cannot be overlooked?

I need to update the test runner to make it fail for “new warnings for `-valid.html`” cases. But no issue needed.

> And, should we reopen #813?

We can’t re-open that. Instead, we need a new PR, but this time either with `-haswarn` renames made to the “problem” tests sources, or else with changes made to the Java code to prevent the warnings for those cases. I mean, if they actually are valid tests and should not be causing warnings.

sideshowbarker:
> > Your call on that but is there or should there be a ticket to make sure these tests cannot be overlooked?
> 
> I need to update the test runner to make it fail for “new warnings for `-valid.html`” cases. But no issue needed.

Fixed that part now.

j9t:
> We can’t re-open that. Instead, we need a new PR, but this time either with `-haswarn` renames made to the “problem” tests sources, or else with changes made to the Java code to prevent the warnings for those cases. I mean, if they actually are valid tests and should not be causing warnings.

Do you like to take care of that? Would otherwise raise a hand but I get the idea that with this situation, that’s the better call? (I can probably pick up a separate issue again shortly.)

PR #1989, opened 2025-12-19
Clarify error message for table body start tags

j9t:
Resolves validator/htmlparser#102

(@sideshowbarker, I think I have all the local setup in place but bear with me if this needs further tweaking 🙂)


sideshowbarker:
This’ll need to be opened instead as PR in the https://github.com/validator/htmlparser — with a patch against those sources.

That said, I’m personally not super enthusiastic about this change being all that necessary. However, the code owners for the parser code may feel differently, I dunno.

Issue #1574, opened 2023-04-05
Hyperlink in error message in web UI goes to wrong line (fragment ID doesn’t match line number it actually links to)


Sivustonikkari:
at https://vapaa-ajattelijat.fi the validation gives the following error:

```
Error: Duplicate attribute src.
At line 900, column 135
ljett.jpg"  src="https://vapaa
```
But when looking at the source, there is only on src element.

sideshowbarker:
The `img` element in the source at line 900 does have a duplicate `src` attribute — the problem is that the https://validator.w3.org/nu/?showsource=yes&doc=https%3A%2F%2Fvapaa-ajattelijat.fi%2F#cl900c135 link in the web UI unexpectedly goes instead to the `img` element at line 937.

In other words, the checker is reporting the correct line number in the error message, and has that same line number int the fragment part of the hyperlink URL in the error message — but that fragment ID doesn’t match the line number it actually links to.

j9t:
Does this issue still persist?

When I’m checking https://validator.w3.org/nu/?showsource=yes&doc=https%3A%2F%2Fvapaa-ajattelijat.fi%2F, the line numbers and links seem to all match up.

Sivustonikkari:
> Does this issue still persist?
> 
> When I’m checking https://validator.w3.org/nu/?showsource=yes&doc=https%3A%2F%2Fvapaa-ajattelijat.fi%2F, the line numbers and links seem to all match up.

Seems to be fine now.

j9t:
Thanks! Closing this one but let’s reopen if this reappears.

FYI @sideshowbarker

PR #2013, opened 2025-12-30
Correct attribute ID generation logic

j9t:
Suspected fix for #1574, but [confirming](https://github.com/validator/validator/issues/1574#issuecomment-3698805505) after issue couldn’t be reproduced anymore.


j9t:
Couldn’t reproduce and therefore validate suspected fix; seems to be confirmed as per https://github.com/validator/validator/issues/1574#issuecomment-3699037756.

PR #2032, opened 2026-01-26
Add warning for `rb` element

### LINKED ISSUE #2031: The `rb` element
### Spec URL

https://www.w3.org/TR/html-ruby-extensions/

### Test case

```html
<ruby><rb>雨</rb><rt>あめ</rt></ruby>
```

j9t:
Resolves #2031

### DIFF
--- a/schema/.drivers/legacy.rnc
+++ b/schema/.drivers/legacy.rnc
@@ -179,13 +179,6 @@ datatypes w = "http://whattf.org/datatype-draft"
 
 	common.elem.phrasing |= plaintext.elem
 
-## Ruby Base: <rb>
-
-	rb.elem =
-		element rb { common.inner.anything & common.attr.anything }
-
-	common.elem.phrasing |= rb.elem
-
 ## Ruby Text Container: <rtc>
 
 	rtc.elem =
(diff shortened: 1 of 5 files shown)


j9t:
@sideshowbarker, just in case, the Claude Code Review check doesn’t seem to be set up fully yet. Known issue?

sideshowbarker:
@claude, code review for the patch in this branch

sideshowbarker:
> @sideshowbarker, just in case, the Claude Code Review check doesn’t seem to be set up fully yet. Known issue?

Dunno. Never used it yet and I don't really know what it's supposed to do by default. I had thought that all it did was just let you @ claude with some prompt

j9t:
Updated the PR given the new information—hope I didn’t miss something.

I can’t find anything on why the Java 25 job would be failing—do you know more?

Not sure about [the review comment about removing this from obsolete elements](https://github.com/validator/validator/pull/2032/files#r2727564382)—probably referred to the other section, in which case this should have been addressed.

As for the text, this was a reference to the Ruby spec and did not need a critique. With this update, there’s currently no special text—if we need any info text, I can add it in this PR or we could add something separately. Let me know.

Issue #2073, opened 2026-04-15
Validator crashes on startup with StackOverflowError


jribbens:
### Summary

I recently upgraded `vnu-jar` from 26.3.10 to 26.4.11, and it now nearly always (but not 100%) crashes immediately (i.e. before even being asked to validate any files) when run in servlet mode.

### URL for document that reproduces the problem

n/a

### HTML/CSS/SVG snippet that reproduces the problem

```html
n/a
```

### Checker error messages

```shell
Exception in thread "main" java.lang.StackOverflowError
at nu.validator.vendor.thaiopensource.relaxng.impl.DuplicateAttributeDetector.startChoice(Unknown Source)
at nu.validator.vendor.thaiopensource.relaxng.impl.ChoicePattern.checkRestrictions(Unknown Source)
  [the above line repeated around 85 times]
at nu.validator.vendor.thaiopensource.relaxng.impl.ChoicePattern.checkRestri / nu.validator.servlet.VerifierServletTransaction - Starting static initializer.
 nu.validator.servlet.VerifierServletTransaction - Reading miscellaneous properties.
 nu.validator.servlet.VerifierServletTransaction - Optional vnu.properties file not found.
 nu.validator.servlet.VerifierServletTransaction - Starting to loop over config file lines.
 nu.validator.servlet.VerifierServletTransaction - Finished reading config.
 nu.validator.servlet.VerifierServletTransaction - Converted config to arrays.
 nu.validator.servlet.VerifierServletTransaction - Prepared namespace array.
 nu.validator.servlet.VerifierServletTransaction - Parsed doctype numbers into ints.
 nu.validator.servlet.VerifierServletTransaction - The cache path prefix is: null
 nu.validator.servlet.VerifierServletTransaction - Parsing set up. Starting to read schemas.
 nu.validator.servlet.VerifierServletTransaction - Will load schema: http://s.validator.nu/html5.rnc
```

### Steps to reproduce

1. `java -cp vnu.jar -Dnu.validator.servlet.bind-address=127.0.0.1 -Dnu.validator.servlet.max-file-size=2097152 nu.validator.servlet.Main 1234`

### Expected behavior

Not crashing.

### Actual behavior

Crashing.

### Screenshots or screen recordings

```
$ java --version
openjdk 11.0.30 2026-01-20 LTS
OpenJDK Runtime Environment Corretto-11.0.30.7.1 (build 11.0.30+7-LTS)
[...]

PR #2089, opened 2026-05-01
Upgrade to jetty 12.0

tenzap:
jetty 11 is not supported anymore upstream

Jetty 12 requires Java/JVM 17 as per https://jetty.org/download.html. So the CI job running with java 11 would fail.

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -127,7 +127,7 @@ DEALINGS IN THE SOFTWARE.
              classpath="${depsDir}/maven-ant-tasks-${maven-ant-tasks-version}.jar"/>
   </target>
 
-  <property name="jetty-version" value="11.0.20" />
+  <property name="jetty-version" value="12.0.34" />
   <target name="dl-deps-jetty" unless="${offline}">
     <echo>Downloading dependencies (jetty)</echo>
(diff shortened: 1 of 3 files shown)


sideshowbarker:
Yeah so it seems like we’re approaching the time when we probably need to quit supporting Java 11.

But @jribbens, given https://github.com/validator/validator/issues/2073 I assume you currently have some environment you need to run the Jetty-based checker in that’s stuck on Java 11. How much heartburn would it cause if that quit working? Are you not able to upgrade to Java 17 in that environment any time soon?

jribbens:
> Yeah so it seems like we’re approaching the time when we probably need to quit supporting Java 11.
> 
> But @jribbens, given #2073 I assume you currently have some environment you need to run the Jetty-based checker in that’s stuck on Java 11. How much heartburn would it cause if that quit working? Are you not able to upgrade to Java 17 in that environment any time soon?

I'm using Ubuntu 24, Amazon Linux 2023, and Windows 11. I doubt any of them are stuck on Java 11, but I think what I read a couple of years ago was that was the version to use for some reason (possibly it was the latest widely-available version?). I find the "Write once, run nowhere" nature of Java very difficult to be honest. But anyway, you probably don't need to worry about me (thanks for doing so!), it looks like Ubuntu 24 and Windows 11 have Java 21, and Amazon Linux has Java 24 available.

sideshowbarker (approved):

PR #2108, opened 2026-06-07
Correct typos, simplify language

j9t:
@sideshowbarker, this fell out of the bag during a larger refactoring:

* fixes some typos as with “the the”
* simplifies language as with “whether or not” (“whether”)
* removes superfluous trailing spaces

Submitting as is as it’s so low-cost, though suggest not to bother if this takes much time or creates a hiccup.

### DIFF
--- a/css-validator/api.html
+++ b/css-validator/api.html
@@ -33,7 +33,7 @@ SOAP 1.2 validation interface documentation</h2>
 <div id="toc">
 <ul>
   <li><a href="#requestformat">Validation Request Format</a></li>
-  <li><a href="#soap12format">SOAP format description</a> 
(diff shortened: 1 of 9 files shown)


sideshowbarker:
Thanks, but all of the files the patch here touches are essentially stuff we’ve vendored in. So while I do think there’s value for us in making these kind of cleanups to source we’ve authored ourselves, I don’t think we want to get started with trying do clean up on stuff that we’ve just vendored in. It’s mostly old stuff — and there’s a lot of other messiness in it that’s not just in the comments.

j9t:
Fair enough—I accepted that risk when preparing.

